PR #7, opened 2019-08-11
Add rudimentary CMake build system

qyot27:
Currently can build tsmuxer CLI and libmediation in native 64-bit on Linux (as verified by `file tsmuxer`).  Windows, Mac, and 32-bit untested (32-bit is mostly a function of the user's CXX_FLAGS settings anyway).

As a consequence of trying to get it to build the correct files, the textSubtitleRender{FT|Win32}.* files had to be moved to an osdep/ subdirectory.  Some #include cases were also fixed to point to the correct headers and directories in the source tree or elsewhere.

Install rules need to be fleshed out for static/shared libs and headers for libmediation, uninstall rules need to be added.

tsMuxerGUI needs to be hooked in as well.

Obviously this is just a starting point and can be tweaked as necessary.  I also wasn't sure what minimum CMake version to set it on, so I just picked one that was still a few years old while still being in the 3.x series.


justdan96:
I'll merge the ARM pull request first. Can you please suggest how the readme should be changed to use CMake instead?

qyot27:
Off the top of my head, `cmake` and `ninja-build` added to the apt-get list.  I'd personally recommend doing out-of-tree builds when using CMake, so something like the following:

```
cd tsMuxer
mkdir tsmuxer-build
cd tsmuxer-build
cmake ../ -G Ninja
ninja
    sudo checkinstall \
    --pkgname=tsmuxer \
    --pkgversion="1:$(./tsmuxer | \
                      grep "tsMuxeR version" | \
                      cut -f3 -d " " | \
                      sed 's/.$//')-git-$(\
                      git rev-parse --short HEAD)-$(\
                      date --rfc-3339=date | sed 's/-//g')" \
    --backup=no \
    --deldoc=yes \
    --delspec=yes \
    --deldesc=yes \
    --strip=yes \
    --stripso=yes \
    --addso=yes \
    --fstrans=no \
    --default ninja install
```
(since the default should be installing it into /usr/local as expected; but at least now it's possible for the user to use `-DCMAKE_INSTALL_PREFIX=/path/to/install/location` so they can output to a different location)

For that matter, CMake is capable of outputting NMake Makefiles, Visual Studio projects, and regular Unix or MinGW Makefiles, and those steps are all *mostly* the same (except for the ones for VS; that requires using `cmake . --build Release` or some such instead of the `make`/`nmake`/`ninja` step).

Due to some of the files needing to be moved in this commit, there would need to be some path changes to the regular Makefiles or project files (and testing to make sure, e.g. the `#include` changes didn't break things) if you want to keep those around for a while before the CMake method stabilizes.  This is currently more of a proof-of-concept (as mentioned, it currently doesn't build tsMuxerGUI), so I wouldn't consider it ready for merge just yet anyway.

If doing cross-compiles, things get a little more complex.  `-DCMAKE_TOOLCHAIN_FILE=/path/to/toolchain/file` and a toolchain file defining the cross-compiler are necessary, but I'd hazard a guess that most distros don't ship a multilib MinGW/GCC compiler the way I have [my mpv build guide set up](https://github.com/qyot27/mpv/blob/extra-new/DOCS/crosscompile-mingw-tedious.txt), so multilib cross-compiles would be way beyond the scope of the readme here.  The actual CMake toolchain file part of those instructions (around line 308) should be fine, though - doing i686 native would require changing the tools' prefixes, but that's all.

justdan96:
I'm just testing out [dockcross](https://github.com/dockcross/dockcross) which I think will solve the issues around toolchains. I'll merge this and then do some further testing.

Issue #9, opened 2019-09-16
Code is not 64-bit clean


blenheimears:
A 64-bit build of tsMuxeR currently creates corrupt but still playable output. For example, the audio and/or video will have glitches. The issue appears to be in the demuxing code rather than the muxing code, as the output is correct if tsMuxeR is given raw elementary streams (eg. .h264, .ac3, .wav files). Additionally, subtitle rendering does not work at all on 64-bit builds. I have determined that this is due to a failure to convert the name of the font to UTF-32, only the first character is converted. I have not determined why the string fails to convert.

filler56789:
Hi there. PERHAPS this post on Doom9's forum has the solution (or workaround) to the problem,
I think it's worth a try:

https://forum.doom9.org/showthread.php?p=1882862#post1882862

justdan96:
My understanding is the 64-bit code won't compile without "-fpermissive" - but perhaps this is actually masking the issues you are describing. I'll try to document the errors that appear if you compile without "-fpermissive" to see if they correspond to the issues you describe.

The issue must be in utf8Converter.cpp - is the build you are using Windows or Linux? Also if you could include some instructions to reproduce the issue with some open source video like https://peach.blender.org/ or http://www.valkaama.com/index.php?page=sources that would really help with the process.

filler56789:
@justdan96 — yeah, the file bitStream.cpp should be rewritten for good. According to what I have found on the web, that kind of conversion [under "normal" circumstances] is not tolerated by the 64-bit compiler(s) because of the "loss of precision" thing. But I myself am not qualified for the task.

blenheimears:
I don't have any issues compiling the code on a 64-bit machine, even without -fpermissive, but the resulting binary does not work correctly. I am using Linux, specifically Ubuntu 18.04. I'll post an example of the corrupt output for Big Buck Bunny tomorrow.

blenheimears:
I've uploaded a file (the first two minutes of Big Buck Bunny) that will trigger the audio bug when processed with a 64-bit build of tsMuxeR. I have also included in the archive the corrupted output of the 64-bit tsMuxer, as well as the correct output of a 32-bit build. I used the m2ts muxing option to create these files, but the bug occurs in the other modes as well, such as Blu-ray ISO. This is Blu-ray quality HD video, so each file is slightly over 500 MB in size once unpacked from the archive.

bbb-2mins.mkv - The original file
[...]

PR #12, opened 2019-10-21
Fix build with Clang

xavery:
Forward declarations are not allowed to have nested namespace declarations, as
they have to appear inside a namespace if the thing they're forwarding appears
in a namespace originally.

This is a warning with GCC, but an error with Clang. Removing the declaration
entirely is the way to go, since psgStreamReader.h includes textSubtitles.h
anyway, which contains a full definition of the TextToPGSConverter class.

### DIFF
--- a/tsMuxer/src/psgStreamReader.h
+++ b/tsMuxer/src/psgStreamReader.h
@@ -9,8 +9,6 @@
 #include "textSubtitlesRender.h"
 #include "textSubtitles.h"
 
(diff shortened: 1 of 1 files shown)


justdan96:
This is great, I've been struggling with getting a build going with osxcross and that also requires Clang. Can you please test your build for the issues given in the other pull request? Preferably using the same files in the Google Drive. I'm curious if you see the exact same corruption issues.

xavery:
I left a comment in #9.

PR #13, opened 2019-10-21
Translate comments from Russian to English

xavery:
CVS log messages are left untouched in files which contain them, and those files
have been converted from Windows-1251 to UTF-8.


justdan96:
Excellent - thanks a lot!

PR #15, opened 2019-10-24
Enhancement crossbuild

justdan96:
switch to cmake. Will gradually feed in any changes required by the Docker builds through this branch.


xavery:
There's generally some more issues related to the CMakeLists.txt changes, mostly related to support for MSVC :

- "Native Windows" platforms don't have pkg-config, which means that compiling outside of MSYS2 under Windows will cause CMake to fail due to being unable to find pkg-config. Unfortunately, I'm not aware of a similar facility for Windows, and any CMakeLists I've written before simply required the user running the build on Windows to manually provide the values for xyz_LIBRARIES and all the other variables automagically provided when running pkg-config.
- Any reason why pkg-config is used for FreeType2, but not for zlib and libpng? The library installation directories, and the names of the library files themselves, can vary across distributions, and using pkg-config is generally the way to go when it comes to smoothing out these differences.
- The file uses compiler switches which are valid for GCC and Clang, but not for MSVC. `-std=c++11` is one of them, but `-static-libgcc -static-libstdc++ -static` is another example - MSVC will either error out or issue a warning about unrecognised compiler flags. I don't do a lot of Windows development, but I think the MSVC equivalents would be [`/MD, /MT, /LD`](https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2019).

I haven't had to make a 100% statically linked executable in a while, so I'm not the best person to know, but I presume there are CMake variables available for doing just that. The general idea of CMake is, after all, to configure as much as possible using its own variables (which are then translated into the proper compiler flags according to the target compiler) and use compiler-specific flags only when there's really no other choice.

justdan96:
Given the issues I've had with MSVC I wasn't planning to support it long-term. However if you have instructions for providing the libraries with Windows it would make sense to merge them.

Probably my own lack of understanding, if you can provide the options I'll merge that as well. 

We can edit the switch for MSVC specifically to use /MD /MT /LD if that is the right way to go.

I am not anti-MSVC it is just that my plan for the long term was using Docker containers for creating the builds for Windows, Linux and Mac and that would mean Msys over MSVC.

xavery:
It's not like it's a priority for me, either - I just assumed you wanted to maintain support for it, given how `if (MSVC)` directives still appear there. :-) I won't spend any more time worrying over that in this case, but I'll keep this in mind as potential future work.

filler56789:
FWIW, the CMakeLists.txt from x265 could give you some "inspiration" (so to speak).
As long as you don't repeat their errors, such as making the newest versions of CMake complain about the "OLD" policies...

https://bitbucket.org/multicoreware/x265/src/default/source/CMakeLists.txt

filler56789:
> "Kudos for saying goodbye to those Makefiles, though."

@xavery — you know that CMake does create makefiles from CMakeLists.txt, don't you?🙂

xavery:
> > "Kudos for saying goodbye to those Makefiles, though."
> 
> @xavery — you know that CMake does create makefiles from CMakeLists.txt, don't you?slightly_smiling_face

Not for me. I use Ninja as the generator, and the main advantage of CMake is that it supports multiple generators, including MSVC project/solution files - which is why I started the (moot now) discussion about MSVC in the first place. :slightly_smiling_face:

justdan96:
I'll take a look at the x265 CMakeLists.txt for some pointers, thanks @filler56789.

I'm so thankful for the community improving this project, I was feeling for a while like I was too busy at work to dedicate time to it but the last few days have been great with the pull requests and comments.

A big thank you to @xavery. I really need to update the README with everyone who's contributed pull requests!

qyot27:
To be honest, the MSVC parts (or, catching both it and MINGW) were sort of bare-minimum affairs to keep things separated out, they weren't really tested and were mostly just stubs for someone else to pick up and refine later on.  When I whipped up the CMakeLists.txt I was primarily dealing with Linux and maybe (I can't really remember) MSys2.

`CMAKE_EXE_LINKER_FLAGS` should not be forcing libgcc or libstdc++ as static.  Especially not on non-Windows, but even on Windows that should be left up to the discretion of the user doing the build.  pkg-config can handle static linking via its own `--static` parameter, which IIRC is why libpng is not detected by the pkg-config routines in CMakeLists.txt: libpng is already specified in Freetype2's pkg-config file as a dependency.  As far as zlib goes, I think that may have just been because it wasn't troublesome to link against if it's present on the system, unlike the way freetype has a reputation of being.  Again, that was 2½ months ago and I can't recall the exact reason.

justdan96:
I can't actually remember why I was forcing static builds so I'm fine with removing that.

PR #16, opened 2019-10-24
Remove condvar, mutex and time from libmediation

xavery:
The title pretty much says it all. This commit replaces the custom wrappers for these three things from libmediation with the ones provided by C++11.


justdan96:
THANK YOU! This is fantastic, I was struggling with this for a while and you've made it look easy!

xavery:
Sadly, this was the stuff I found the easiest to remove : I don't think replacing `TerminatableThread` (for instance) will be as quick due to the fact that you can't really inherit from `std::thread`, so it will require much more restructuring. I'll probably get to it someday.

PR #17, opened 2019-10-25
Enhancement crossbuild

justdan96:
I realised I was getting weird errors building in a Docker container based on dockcross - so I just installed MXE directly into my Ubuntu VM and started testing the builds that way.

This branch includes some fixes to the CMake configuration, fixes for Windows builds and some changes to the README.


CMakeLists.txt
```diff
@@ -31,10 +30,18 @@ add_executable (tsmuxer ${tsmuxer_sources})
...
 find_package (PkgConfig REQUIRED)
+
+# freetype2 as a dependency
 pkg_check_modules (FREETYPE2 REQUIRED freetype2)
 
+# zlib as a dependency
+pkg_check_modules (ZLIB REQUIRED zlib)
+
+#libpng as a dependency
+pkg_check_modules (LIBPNG libpng REQUIRED)
```
xavery: libpng is only needed by freetype2 when building statically, and as such it doesn't make much sense to always link against it. Relevant pkg-config calls :
`$ pkg-config --libs freetype2
-lfreetype `
`$ pkg-config --libs --static freetype2
-lfreetype -lbz2 -lz -lpng16 -lm -lm -lz -lharfbuzz -lm -lglib-2.0 -pthread -lpcre -pthread -lgraphite2 `
This is handled by my recent changes to CMakeLists.txt, so you can delete these lines when merging.

justdan96:
this now just has the small code changes required by MXE and the updated readme

PR #18, opened 2019-10-25
Completely rework the CMake build system

xavery:
So this is a rather large change, and as such warrants a sufficiently large description. Here's a rundown of the changes :

- The whole CMake build system has been reworked. The `CMakeLists.txt` file in the main directory of the project is now just an aggregator for "per-project" `CMakeLists.txt` files which reside in individual subdirectories, and sets flags which for all subprojects.
- The user can now choose between a statically- and dynamically- linked build of the main tsmuxer executable with the variable `TSMUXER_STATIC_BUILD`, which defaults to false.
- Useless libmediation defines have been removed : `LIBMEDIATION_API` was probably meant to be defined to something related to symbol visibility (`__declspec(dllexport)` on Windows and the visibility attribute on ELF platforms), but it doesn't make sense if it's built as a static library. The rest (`DISABLE_LOG_SUPPORT`, `DEBUG_NEW12` and `NO_SYSTEM_EXCEPT`) were totally unused.
- Unused source files in tsmuxer (`vodCoreContext.cpp` and `vodTransport.cpp`) were moved to an `unused` directory.
- tsmuxer source files now reside directly in the tsmuxer directory instead of an extra `src` subdirectory.
- External dependencies are now always queried by `pkg-config`.
- The QMake .pro file used for building tsMuxerGUI was replaced with an equivalent `CMakeLists.txt`.
- The Visual Studio project/solution files were removed, as they can now be generated by CMake directly if need be.
- Mistakenly committed auto-generated Qt files (`ui_*.h`) were removed.

I'm still not entirely satisfied with how the dependency between tsmuxer and libmediation is expressed, but seeing how libmediation is probably going away sooner or later, I don't think it's necessary to worry about it too much...


justdan96:
Excellent - we are definitely making good progress!

I'll merge this one and then check for conflicts against the other pull request.

xavery:
Well, that was much faster than I had expected. ;-)

filler56789:
@xavery — do you have an E.T.A. for fixing the file **bitStream.cpp** so that the 64-bit GCC doesn't require the _-fpermissive_ flag?

Current line 53 still is:

`for (; len >= 8 && ((unsigned long) src % sizeof(unsigned)) != 0; len -= 8) {`

which makes the compiler complain about the "loss of precision" in this type of conversion.

The fix would simplify the main CMakeLists.txt, as a consequence.

xavery:
@filler56789  Actually, I didn't have any problems related to compiling with neither GCC nor Clang on Linux, even before introducing `-fpermissive`, which is why I didn't really spend a lot of time trying to fix the underlying problem.

The only platforms where this would cause a loss of precision would be those where `unsigned long` isn't equal to the size of a pointer. Those would be platforms using the LLP64 data model (64-bit Windows at the time of writing), but it still doesn't explain why GCC would issue this error, unless it occurred when compiling with MSYS2 for a 64-bit target, which is the only possibility I see. While casting a pointer to an integer type might (and most of the time does) signal something weird going on, it has its uses and I'm certain that there's a reason why it's used here. Fortunately, C++11 introduced the `std::uintptr_t` type, which is an "unsigned integer type capable of holding a pointer" - which I *presume* was what the original author meant to express here.

However, `moveBits` is totally unused and if you remove it, the whole program compiles okay. I guess there's no real point in keeping it. On the other hand, if you have a setup where the problem is reproducible (i.e. the compiler actually refuses to compile the file without `-fpermissive`), you can try changing this line to :

```
for (; len >= 8 && (reinterpret_cast<std::uintptr_t>(src) % sizeof(unsigned)) != 0; len -= 8) {
```

And add a `#include <cstdint>` at the beginning to be sure that `std::uintptr_t` is actually defined.

Let me know if that works out.

filler56789:
@xavery — Many thanks for the detailed answer. Yes, the problem happens while using MSYS2 + GCC + Win64. But only under, let's say, "certain circumstances". Here goes the link to my post on Doom9's forum:

https://forum.doom9.org/showthread.php?p=1882862#post1882862

I will give a try to your suggestion (_reinterpret_cast... + #include..._) A.S.A.P.

filler56789:
Just confirming that your recipe worked 🙂

I have updated my _RV-tsMuxeR_ repository accordingly.

Again, many thanks.

xavery:
Fantastic! I'll send a pull request for this change, then.

PR #20, opened 2019-10-26
Fix code which needed -fpermissive on LLP64 platforms

xavery:
Tested by @filler56789 in #18


justdan96:
I was going to post this as an issue but you're too fast! Great work once again!

PR #21, opened 2019-10-26
Use std::thread in libmediation's TerminatableThread

xavery:
TerminatableThread is now implemented by using std::thread. The terminate()
function provided by the base class was actually not used anywhere, and was
removed along with the base class itself.
As a bonus, the completely unused files common_win32.{cpp,h} are also removed.


justdan96:
Great stuff!

PR #22, opened 2019-10-26
Remove stdafx.{cpp,h}

xavery:
Precompiled headers aren't actually used by any of the compilers, so leaving
this header in the repo doesn't make much sense.
However, removing it revealed some places in the code which rely on windows.h
being included on WIN32, which was probably included via the tchar.h include.
These places have been fixed to explicitly include windows.h.


justdan96:
I'll test this against my MXE installation but it looks like an excellent addition!

PR #23, opened 2019-10-27
Fix MSYS2 build

xavery:
The MSYS2 build failed due to using stuff from `string.h`, but the header not being included. This commit fixes this.


justdan96:
You're a star!

Issue #27, opened 2019-11-19
Synchronisation Issues with UHD HEVC


justdan96:
From a5180007 here: https://forum.doom9.org/showthread.php?p=1888650#post1888650

There is a bug creating unsynchronized AV with some HEVC files, due to insufficient size of coded picture buffer.
In brief, when reading a nal the end of the buffer is reached before finding the next nal start_code, and the PTS is not incremented for this nal.

I have corrected the bug on my side by simply changing the line 7 in avPacket.h from:
```
const static int MAX_AV_PACKET_SIZE = 32768;
```
to
```
const static int MAX_AV_PACKET_SIZE = 800000;
```

For testing the issue HEVC from https://uhdsample.com are a good choice. I've tried "Samsung 7 Wonders Of The World UHD 4k Sample 60 fps", with tsMuxer 2.6.12 it muxes 13,303 frames, and by increasing max frame size to 800,000 it muxes 13,773 frames: 470 HEVC frames (nearly five seconds) had been discarded by 2.6.12. You might have not seen it as nobody speaks in the video so it it difficult to spot audio delay.

However this video really has 13823 frames, so there are still 50 frames missing : the video is not bluray compliant, the frames are single slice (i.e. not split in four slices) and the biggest frame size in the video is 1,565,299 bytes (frame 7,106), which shows that increasing max frame size to 800,000 is not enough for non blu-ray compliant UHD. As stated previously, the permanent fix should be to rewrite code to get rid of max frame limitation.

justdan96:
It's a silly solution for now but I'm tempted to see what happens with RAM usage and general performance if we change the type of MAX_AV_PACKET_SIZE to an uint32_t and change the value to UINT32_MAX.

PR #29, opened 2019-11-24
Get rid of MAX_AV_PACKET_SIZE limitation

jcdr428:
Should solve issue 27.

tsMuxeR limits the size of Audio Video frames to 32768 bytes (value of MAX_AV_PACKET_SIZE in avPacket.h).

This patch:
- forces to check the size of next NAL for all video frames -and not only when remaining size of buffer is below MAX_AV_PACKET_SIZE as previous.
- removes truncating of last frame.


justdan96:
Resolves #27

Issue #26, opened 2019-11-19
Cannot handle AAC frame size above 2048


justdan96:
Reported by a5180007 here: https://forum.doom9.org/showthread.php?p=1890608#post1890608

Another bug for aac muxing. In brief, the aac frame size is written on 13 bits (bits 31-43 in the frame), however tsMuxeR takes into account only the first 11 bits. So when the aac frame size is above 2048 bytes, tsMuxeR incorrectly reports a bad frame and discards it.

In aac.cpp, replace line 52:
```
int AACCodec::getFrameSize(uint8_t* buffer)
{
	return buffer[4]*8 + (buffer[5] >> 5);
}
```
with
```
int AACCodec::getFrameSize(uint8_t* buffer)
{
	return ((buffer[3] & 0x03) << 11) + (buffer[4] << 3) + (buffer[5] >> 5);
}
```

Any 5.1 channel AAC audio should trigger the bug. (File shared privately, public example to be released soon!)

PR #30, opened 2019-11-24
Correct AAC frame size parsing

jcdr428:
This patch solves issue 26.

tsMuxeR incorrectly parses the AAC frame length : as per ISO/IEC 13818-7, frame_length is 13-bit, however tsMuxer reads 11 bits only. Therefore all frames above 2048 bytes are reported as bad frames and discarded.


justdan96:
Resolves #26

PR #51, opened 2019-11-30
Correct aac frameSize

jcdr428:


justdan96:
Looks like a small change, will merge this one.

PR #52, opened 2019-11-30
Corrected conflict beween min and std::min

jcdr428:
This patch corrects the conflit between <algorithm> std::min and <windows.h> min classes on Windows.
Alternative would be to use #define NOMINMAX

[Sorry, the commits are a little messy, I'm learning how to merge the commits on github.com]


justdan96:
Can you please let me know what the behaviour is before and after this change? I just want to be sure on what exactly this fixes.

jcdr428:
@justdan96, here is the build error log I get on Windows when the std::min is not in brackets:
[Error log - min.txt](https://github.com/justdan96/tsMuxer/files/3908796/Error.log.-.min.txt)

See https://stackoverflow.com/questions/5004858/why-is-stdmin-failing-when-windows-h-is-included

justdan96:
What's the advantage of this over #define NOMINMAX ?

jcdr428:
I read potential problems reported with the #define NOMINMAX, e.g.
https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max/13420838
https://stackoverflow.com/questions/15900381/nominmax-with-visual-studio-2012-mfc-project

It seemed to me that simply putting std::min under brackets was easier and garanteed to be hassle free on all platforms.

justdan96:
Fair enough, I'll resolve the merge conflicts and get this merged.

justdan96:
Merged in #56

PR #53, opened 2019-11-30
WIN32_LEAN_AND_MEAN for Windows built

jcdr428:
On MSVC 16.3.10, textSubtitles.cpp won't compile without this line (which was originally in the removed file stdafx.h). See attached Error log without this line.
[Error Log.txt](https://github.com/justdan96/tsMuxer/files/3907831/Error.Log.txt)


justdan96:
Seems reasonable and doesn't seem to affect the other builds

PR #54, opened 2019-12-01
UHD Fix to tsMuxer.cpp

jcdr428:
Fix to take into account change of UHD option from --blu-ray v3 to blu-ray-v3


justdan96:
No issue with this one, good catch!

PR #57, opened 2019-12-02
UHD Fix to muxerManager.cpp

jcdr428:
Patch needed due to change from "--blu-ray v3" to "--blu-ray-v3"
(same as patch for tsMuxer.cpp already merged)


justdan96:
Another good catch - thank you!

PR #58, opened 2019-12-02
Parse hdr10 metadata

jcdr428:
Parse HDR10 metadata from HEVC stream, by detecting and parsing SEI NAL units with Payload 137 and 144.
This data will be used to inject in mpls/clpi/bdmv files (future patches)


jcdr428:
@justdan96 I am about to complete the patch with addition of detection of Dolby Vision (nal_unit_type 6201 or 6301).
So I close this push and will resubmit when complete.

PR #62, opened 2019-12-06
Fix conflict between std::max and windows.h max

jcdr428:
Same as previous fix for std::min...


justdan96:
Makes sense to me, happy to get this merged.

PR #65, opened 2019-12-07
Make the GUI look for the main binary under "tsmuxer" and "tsMuxeR"

### LINKED ISSUE #63: tsMuxeR not found!
Hello,
I have problem when I want to add input file I received error:

Manjaro repo arch
N60xHu1

xavery:
The all-lowercase binary name is the one that's emitted by the current build
system, while the old one is kept for backwards compatibility.

Fixes #63


xavery:
There are lots of seemingly redundant whitespace changes in the header file, but it had a mixture of CRLF and LF line endings : this is now normalised to LF.

PR #67, opened 2019-12-08
Replace "virtual" with "override" where a given function is overridden

xavery:
With C++14 being officially enabled for some time now, all those defines turning
"override" into an empty token looked kind of embarassing. They are now removed
and the "virtual" keyword is replaced with "override" where appropriate.

### DIFF
--- a/libmediation/containers/safequeue.h
+++ b/libmediation/containers/safequeue.h
@@ -117,7 +117,7 @@ public:
 	{
 	}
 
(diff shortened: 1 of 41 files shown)


justdan96:
Merged as this: https://github.com/justdan96/tsMuxer/commit/e575bc0f806dfaf0ee5025103afd56f56884b650

PR #70, opened 2019-12-10
Patch for UHD HDR10 and HDR10+

jcdr428:
I've tested on few SDR and HDR10 Blu-rays without issues.

The patch would need to be thoroughly tested by some other users (eg offer an alternative build for testing) before integration to Master.


justdan96:
It looks like these changes conflict with some of the changes @xavery introduced in #67 

Can you please take a look at how to resolve these conflicts?

jcdr428:
Grr... I have to replace most "virtual" with "overrides". Will do !!

jcdr428:
@justdan96 the only conflicts left are the integration of functions getStreamHDR() in the four StreamReaders. Nothing else I can do on my side.

tsMuxer/BufferedReader.h
```diff
@@ -5,6 +5,7 @@
 #include <map>
 #include <types/types.h>
 #include <system/terminatablethread.h>
+#include <mutex>
```
xavery: `safequeue.h` that's included in the following line already includes `<mutex>` - are you sure this is needed here?

tsMuxer/blurayHelper.cpp
```diff
@@ -165,22 +166,29 @@ bool BlurayHelper::writeBluRayFiles(bool usedBlackPL, int mplsNum, int blankNum,
...
     }
     else if (m_dt == UHD_BLURAY) {
         bdMovieObjectData[5] = bdIndexData[5] = '3';
-        fileSize = 0x78;
+        fileSize = 0x9C; // add 36 bytes for HDR data extension
+        bdIndexData[15] = 0x78; // start address of HDR data extension
+        emptyCommand = bdIndexData + 0x78;
+        memcpy(emptyCommand, "\x00\x00\x00\x20\x00\x00\x00\x18\x00\x00\x00\x01"
+                             "\x00\x03\x00\x01\x00\x00\x00\x18\x00\x00\x00\x0C"
+                             "\x00\x00\x00\x08\x51\x00\x00\x00\x00\x00\x00\x00", 36); // HDR data extension
```
xavery: If you replace this string literal with an array of `uint8_t`, you'll get less code and will be able to use `sizeof` instead of hard-coding the size of the array.

tsMuxer/blurayHelper.cpp
```diff
@@ -165,22 +166,29 @@ bool BlurayHelper::writeBluRayFiles(bool usedBlackPL, int mplsNum, int blankNum,
...
+        bdIndexData[0x96] = (byte)HDR10_metadata[0]; // HDR flags
```
xavery: `std::byte` is a C++17 feature, and this code rightly causes a compilation failure with Clang. `static_cast` to `uint8_t` instead.

tsMuxer/abstractStreamReader.h
```diff
@@ -32,6 +32,7 @@ class AbstractStreamReader: public BaseAbstractStreamReader {
 	virtual int readPacket(AVPacket& avPacket) = 0;
 	virtual int flushPacket(AVPacket& avPacket) = 0;
 	virtual int getTSDescriptor(uint8_t* dstBuff) {return 0;}
+	virtual int getStreamHDR() const { return 0; }
```
xavery: This causes `getStreamHDR` not to be abstract in all the classes deriving from `AbstractStreamReader`, notably `MPEGStreamReader` which redeclares the function as abstract. Is this intentional?

tsMuxer/h264StreamReader.h
```diff
@@ -3,6 +3,7 @@
 
 #include <map>
 #include <iostream>
+#include <algorithm>
```
xavery: There is no code in the header which directly uses stuff from `<algorithm>` - why was this moved here from the source file?

tsMuxer/hevc.cpp
```diff
@@ -807,6 +809,74 @@ int HevcPpsUnit::deserialize()
...
+		return rez;
+	try
+	{
+		int payloadType;
+		int payloadSize;
+		int nbyte;
+		do {
+			payloadType = 0;
+			payloadSize = 0;
+			nbyte = 0xFF;
```
xavery: Any reason why these variables are initialised later (unconditionally, since the `do` block is going to be entered either way) instead of at declaration time? If not, just initialise them while declaring.

tsMuxer/hevc.h
```diff
@@ -3,6 +3,8 @@
 
 #include "nalUnits.h"
 
+extern int HDR10_metadata[7];
```
xavery: This is a global variable that's set by parts of `hevc.cpp` and then read by `blurayHelper.cpp`. Would it be possible to reduce its scope, for example to a member variable of the `BlurayHelper` class?

tsMuxer/hevcStreamReader.h
```diff
@@ -41,6 +42,7 @@ class HEVCStreamReader: public MPEGStreamReader
     HevcVpsUnit* m_vps;
     HevcSpsUnit* m_sps;
     HevcPpsUnit* m_pps;
+	HevcSeiUnit* m_sei;
```
xavery: Any reason why it's not a `std::unique_ptr`?

tsMuxer/metaDemuxer.h
```diff
@@ -10,6 +10,8 @@
 #include <set>
 #include <chrono>
 #include <queue>
+#include <iomanip>
+#include <chrono>
```
xavery: Not only is `<chrono>` already included by this header, but there is no code that uses `<iomanip>` which would require its inclusion here.

tsMuxer/blurayHelper.cpp
xavery: Looks like [GCC doesn't like this either](https://github.com/jcdr428/tsMuxer/commit/d8d9669973af16b4e6c64eb6b0d08686327ad099/checks#step:5:62).

tsMuxer/BufferedReader.h
jcdr428: @xavery you are correct. I didn't apply the changes to libmediation, so safequeue.h was still referencing to system/mutex.h

tsMuxer/abstractStreamReader.h
jcdr428: @xavery you're correct: removed from AbstractStreamReader. First experience with C++, I'm still unfamiliar with inheritance...

tsMuxer/hevc.cpp
jcdr428: @xavery once again, so true... Thanks for correcting.

tsMuxer/metaDemuxer.h
jcdr428: Both removed !

tsMuxer/h264StreamReader.h
jcdr428: @xavery This is correct. Reference removed.

tsMuxer/hevc.h
jcdr428: @xavery I've tried for hours to set the variable in a specific class. Then it has to be passed from class to class, I didn't manage to do this and ended up setting the global variable... Maybe someone else will manage to do this.

tsMuxer/hevcStreamReader.h
jcdr428: @xavery not sure what you mean here. I used the same struct for sei as for vps, sps and pps.

tsMuxer/hevc.h
xavery: I understand - sadly, I also have no idea on how to get this through in a more civilised way, so I'll just let it go.

tsMuxer/hevcStreamReader.h
xavery: There is no reason for this to be a raw pointer : since C++11, we can use the `std::unique_ptr` class, which guarantees that the pointer is never "accidentally" copied and its ownership is always tied to the `unique_ptr` object. Existing code doesn't do this, because it was written before these classes were available.

xavery:
@jcdr428 I'm fine with merging these changes once you're confident that enough testing has been done.

jcdr428:
@xavery thanks. jdobbs and I have tested this HDR10 patch quite a lot, and it is working well.
tymoxa has also tested and has reported incorrectly formed TS packets / descriptors which are not linked to this patch -I have another patch ready to be pushed for correcting the HEVC and AVC descriptors.

So in brief, yes I am confident that enough testing has been done to merge this change.

PR #74, opened 2019-12-13
WIP: Allow users to specify explicit PIDs via the meta file

xavery:
According to [this post](https://forum.doom9.org/showthread.php?p=1892798#post1892798) on Doom9, it would be nice if TSMuxer allowed specifying PIDs of the streams that are going to end up in the TS explicitly.

This adds a `pid` parameter to the stream descriptions in the meta file, which allows the users to specify the PID that they want the given stream to have in the TS, as long as it doesn't collide with any predefined ones.

@jcdr428 is working on the real reason why this is needed (Dolby Vision streams need a specific PID), but this doesn't collide with their work - and I hope that the little refactoring I've done turns out to be helpful while developing.

### DIFF
--- a/tsMuxer/tsMuxer.cpp
+++ b/tsMuxer/tsMuxer.cpp
@@ -47,6 +47,21 @@ static const int PAT_PID = 0;
 static const int SIT_PID = 0x1f;
 static const int NULL_PID = 8191;
 
+namespace {
+bool isValidCustomPID(int pid) {
+    switch(pid) {
+    case PAT_PID:
+    case SIT_PID:
+    case DEFAULT_PCR_PID:
+    case DEFAULT_PMT_PID:
+    case NULL_PID:
+        return false;
+    default:
+        return pid < NULL_PID;
+    }
+}
+}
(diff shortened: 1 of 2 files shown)


justdan96:
Great work! Can you give an example invocation? May need to update the documentation.

xavery:
I forgot to mark this PR as WIP - this should be merged once documentation is updated and GUI support is ready. I'll let you know once both are ready.
As for the sample, here's the metafile I've been using while testing :
```
MUXOPT --no-pcr-on-video-pid --new-audio-pes --vbr  --vbv-len=500
V_MPEG4/ISO/AVC, "raw/video.h264", insertSEI, contSPS, pid=1234
A_AC3, "raw/audio01.ac3", pid=0xf0f
```

xavery:
Since @jcdr428 's changes concerning Dolby Vision support are now merged - and some of the stuff I worked on in this PR is now obsolete/conflicting, I'm delaying this until further notice. Will submit a separate PR once the work is done.

PR #88, opened 2019-12-21
Fix numbers appearing in places where the language code should

### LINKED ISSUE #87: Problem adding external .sup and idx/sub
When adding an external .sup-file the chosen language shows as an eight digit number.
idx/sub are not accepted at all.

xavery:
The original code paired the position inside the langComboBox with an address
into the shortLangList/fullLangList arrays by leveraging the second QVariant
argument of QComboBox::addItem. This was achieved by casting the pointer firstly
to a void*, and then to a qulonglong. However, the code using these QVariants
in QComboBox signal handler functions has since been changed to always call
toString(), which was incorrect and led to addresses being shown in the UI in
the form of longish decimal numbers.

Funnily enough, it looks like these shenanigans weren't even really needed : the
only part of the language pair used in the signal handler functions is the
three-letter code, which can be simply converted to a QString at the time
QComboBox::addItem is called.

Fixes #87.


xavery:
@justdan96 I hope you don't mind me self-merging, but I wanted to get this to today's nightly build.

justdan96:
No issue with that!

PR #90, opened 2019-12-27
Automatic V3 for HEVC

jcdr428:
This patch allows tsMuxer to automatically switch to Blu-ray V3 when HEVC is detected.
The -v3 option is still needed to manually select blu-ray V3 version for AVC. However IMO it would be better to select the GUI V3 option in the tab "Blu-ray", rather than the two added UHD radio buttons on the main tab.

### DIFF
--- a/tsMuxer/muxerManager.h
+++ b/tsMuxer/muxerManager.h
@@ -45,7 +45,7 @@ public:
 
     void parseMuxOpt(const std::string& opts);
     int getTrackCnt() { return (int) m_metaDemuxer.getCodecInfo().size();}
-    bool getHevcFound() { return (bool) m_metaDemuxer.m_HevcFound; }
+    bool getHevcFound() { return m_metaDemuxer.m_HevcFound; }
     AbstractMuxer* getMainMuxer();
     AbstractMuxer* getSubMuxer();
     bool isStereoMode() const;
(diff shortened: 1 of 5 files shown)


justdan96:
Great! I can take a quick look at the GUI code in the next few days - or is that in the pipeline?

tsMuxer/muxerManager.h
```diff
@@ -45,7 +45,7 @@ class MuxerManager
 
     void parseMuxOpt(const std::string& opts);
     int getTrackCnt() { return (int) m_metaDemuxer.getCodecInfo().size();}
-
+    bool getHevcFound() { return (bool) m_metaDemuxer.m_HevcFound; }
```
xavery: `m_HevcFound` is already a `bool`. The cast is redundant. Also, this should be a `const` function.

jcdr428: Hi xavery, cast (bool) removed. I don't think it should be a const, the value of m_HevcFound varies with each line of the meta file.

jcdr428:
> Great! I can take a quick look at the GUI code in the next few days - or is that in the pipeline?
@justdan96 I haven't installed Qt5, so if you can make the change in the GUI this would be great.

jcdr428:
@xavery the patch is complete.

PR #92, opened 2019-12-27
Correct TS Descriptors

jcdr428:
Correct TS Descriptors to be the same as the descriptors on commercial Blu-ray disks.


justdan96:
Happy to merge if this has been successfully tested on a few examples - can you please confirm that?

jcdr428:
@justdan96 yes, tested with several commercial Blu-rays, and the AVC and HEVC registration descriptors are now the same.

PR #94, opened 2019-12-29
Solve HEVC Buffer Overflow Bug

jcdr428:
Currently, buffer overflow is not avoided in all cases, i.e. nextNal can have end buffer value instead of start of next nal value.
This results in the AUD NAL not being always at the begining of a PES packet.
This patch solves this by ensuring that when nextNal == bufEnd, data is always added to buffer.

### DIFF
--- a/tsMuxer/hevc.h
+++ b/tsMuxer/hevc.h
@@ -41,7 +41,11 @@ enum HEVCUnitType {
     NAL_FD_NUT     = 38,
     NAL_SEI_PREFIX = 39,
     NAL_SEI_SUFFIX = 40,
-	NAL_DV         = 62,
+    NAL_RSV_NVCL45 = 45,
+    NAL_RSV_NVCL47 = 47,
+    NAL_UNSPEC56   = 56,
+    NAL_DV         = 62,
+    NAL_UNSPEC63   = 63
 };
 
 struct HevcUnit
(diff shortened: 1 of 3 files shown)


justdan96:
Has this been tested with very large files? I think a good test would be to get "Samsung 7 Wonders Of The World 4K Demo" in HEVC format and check the mediainfo for the before and after files.

jcdr428:
@justdan96 thanks for pointing me to "Samsung 7 Wonders Of The World 4K Demo.ts" : actually many frame VCL Nals are followed by FD_NUT (filler data) Nals, which only use is to keep a constant bitrate. The presence of these Nals is not covered by tsMuxer.

So yes, I will have to complete the patch to take into account the filler data Nals. Please put the patch as "wip".

justdan96:
No problem, we really need to come up with a proper test suite at some point, to make sure we are covering all example cases.

jcdr428:
@justdan96 I have tested this fix with several m2ts (Blu-rays and others), it is working well and ready to be merged.

This solves the BD-ROM Part3 Verifier error HEVC0112:

----------------------------------------------
Error ID      : HEVC0112
Target File   : 00000.m2ts <HEVC PID=0x1011>
Target Field  : <PES packet of the HDMV HEVC video stream>
Section No    : 9.19.7
Error Message : The first byte of a PES packet payload is not the first byte of video access unit.
                
                File offset of PES packet = 4057544
                File offset of PES packet payload = 4057558
                File offset of the video access unit =4223220.
Explanation   : The first byte of a PES packet payload shall be the first byte of video access unit. 
                One PES packet shall contain only one access unit.
----------------------------------------------

PR #95, opened 2019-12-29
Revert TS_Recording_Rate to 60 MB

jcdr428:
The coarse entries and fine entries in the .clpi CPI table depend of the TS_Recording_Rate.
Rather than modifying them for UHD, the recording rate is reverted back to 60 MB (same as HD) as tsMuxer should be used mainly with 50 GB Blu-ray Disks.

### DIFF
--- a/tsMuxer/blurayHelper.cpp
+++ b/tsMuxer/blurayHelper.cpp
@@ -178,7 +178,7 @@ bool BlurayHelper::writeBluRayFiles(bool usedBlackPL, int mplsNum, int blankNum,
         emptyCommand = bdIndexData + 0x78;
         memcpy(emptyCommand, "\x00\x00\x00\x20\x00\x00\x00\x18\x00\x00\x00\x01"
                              "\x00\x03\x00\x01\x00\x00\x00\x18\x00\x00\x00\x0C"
-                             "\x00\x00\x00\x08\x51\x00\x00\x00\x00\x00\x00\x00", 36); // HDR data extension
+                             "\x00\x00\x00\x08\x21\x00\x00\x00\x00\x00\x00\x00", 36); // HDR data extension
         bdIndexData[0x96] = (uint8_t)HDR10_metadata[0]; // HDR flags
 
     }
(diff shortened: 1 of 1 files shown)


justdan96:
What is the expected impact here? I just want to make sure we aren't going to interfere with any expected behaviour from the users.

jcdr428:
TSM 2.6.12 was set at BD 50GB, transfer rate 48 MB.
The HDR10 patch has changed it to UHD BD 100 GB, transfer rate 109 Mbps -same as UHD BDs.
See http://www.hughsnews.ca/faqs/authoritative-blu-ray-disc-bd-faq/8-recording-and-reading-speed

Two reasons to revert to former value:
1. The calculation of the coarse and fine entries in the .clpi CPI table should change accordingly; but
2. I understand that UHD BD-ROM and the BD XL do not use the same format: so you cannot burn a Bluray on a BD-R or BD-RW 100 GB and read it on your UHD BD Player.

So currently tsMuxer will be limited to burning to 50 GB BD-R, no point correcting 1... at least until there are some players reading XL BDs.

FilipeAmadeuO:
The coarse entries and fine entries in the .clpi CPI table should be defined by type of disc (HD or UHD) selected when building. 
Please note that some people build in UHD, because they want to replace subtitles/audio and keep video. The goal can be use to play iso or folder directly on some player (ex: Oppo). 
Please do not revert but define clpi entries depending on remuxing for HD or UHD.

jcdr428:
@FilipeAmadeuO you can still build in UHD but you'll have to burn on a BD-50, as the BD-XL are currently not compatible with UHD BD players.

FilipeAmadeuO:
In my case the goal is not burn it to BD-R, but keep it in BD-100 and play as iso file. I’m afraid that this change can create incompatibility issues in a OPPO player because it thinks that’s a BD50 and consequently a HD and not UHD. 
Any sugestion ?

jcdr428:
@FilipeAmadeuO ok, I understand. I'll work on the changes on the CPI table later.
@justdan96 in the meantime IMO it would still be better to push the commit so that the Transfer_Rate is in line with the CPI table. I'll set back the 100 GB BD/109 Mbps wihen I have completed the fix on the CPI table.

justdan96:
I'm wary about pushing anything that changes expected behaviour for end-users, would you be able to include the CPI table changes in this pull request as well?

jcdr428:
@justdan96 the CPI table has been changed and is still in line with the former Transfer Rate of 48 Mbps: this is why I am insisting to revert the TR back to 48 Mbps.

PR #100, opened 2019-12-30
Set video stream number to 1

jcdr428:
As per T-REC.H-222.0 (ISO/IEC 13818-1) Table 2-22, video stream_id is 1110 xxxx for video stream number xxxx.
As checked on multiple commercial Blu-rays, video stream_id should be 1, not 0.

This fix solves error HEVC0116 reported by BD-ROM Part3 Verifier:

    ----------------------------------------------
    Error ID      : HEVC0116
    Target File   : 00000.m2ts <HEVC PID=0x1011>
    Target Field  : <PES packet of the HDMV HEVC video stream>.stream_id
    Section No    : 9.19.7
    Error Message : stream_id is set to 0xe0.
                
                File offset of PES packet = 776.
    Explanation   : stream_id: This field shall be set to 1110 0001b.
    - - - - - - - - - - - - - - - - - - - - - - - 
    [HEVC0116] : 346 times.
    ----------------------------------------------

### DIFF
--- a/tsMuxer/tsPacket.h
+++ b/tsMuxer/tsPacket.h
@@ -27,7 +27,7 @@ static const uint8_t PES_PRIVATE_DATA2        = 0xbf;
 static const uint8_t PROGRAM_STREAM_DIRECTORY = 0xff;
 
 static const uint8_t PES_AUDIO_ID = 0xc0;
-static const uint8_t PES_VIDEO_ID = 0xe0;
+static const uint8_t PES_VIDEO_ID = 0xe1;
 static const uint8_t PES_VC1_ID   = 0xfd;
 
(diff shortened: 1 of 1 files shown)


justdan96:
100 pull requests already! This seems straightforward. 

By the way how does someone get their hands on that Sony BD-ROM Verifier?

jcdr428:
Good question. I'd like to have it as well, this would speed the work on tsMuxeR !
One Doom9 member has kindly tested my builds and sent me the reports :
[testUHDtsMuxeR.zip](https://github.com/justdan96/tsMuxer/files/4012315/testUHDtsMuxeR.zip)

Still some work to be done :)

Issue #104, opened 2020-01-02
First MPLS / M2TS and Chapters


zamengo:
I use the w64-nightly-2020-01-02--01-10-53 and have this errors When use "UHD Blu-ray folder":

- First MPLS dont respect the option after remux.
- First M2TS dont respect the option after remux.
- Chapters dont respect the list, put only "00:00:00.000" after remux.

jcdr428:
@justdan96 @xavery straightforward fix. Could you please merge the push #106 asap for testing (I don't have qt5 installed so can't check whether the issue is fully solved).

justdan96:
#106 has now been merged and will be a part of the next nightly build.

zamengo:
Nice one! Works fine now!

PR #106, opened 2020-01-04
Correct UHD Bluray Folder Behavior

jcdr428:
This fixes e.g. lack of chapters with UHD Bluray Folder option.

(see issue #104 and https://forum.doom9.org/showthread.php?p=1894587#post1894587)


justdan96:
Tested with a local build - no issues.

PR #111, opened 2020-01-06
Dolby Vision Integration

jcdr428:
- Separate the V3 flags (for index.bdmv) from the HDR metadata (for .mpls)
- With V3_flags the UHD_BLURAY DiskType is not needed anymore
- Add subpath stream_id=10 for Dolby Vision
- When 4K is detected, change BD Type from 25/50 GB to 66/100 GB,
   and TS-Recording_Rate from 48 mbps to 109 mbps

This DV patch has been successfully tested by some Doom9 members from my fork.
Cf. https://forum.doom9.org/showthread.php?p=1894442#post1894442 and following posts.

### DIFF
--- a/tsMuxer/blurayHelper.cpp
+++ b/tsMuxer/blurayHelper.cpp
@@ -111,8 +111,7 @@ bool BlurayHelper::open(const string& dst, DiskType dt, int64_t diskSize, int ex
 
 bool BlurayHelper::createBluRayDirs()
 {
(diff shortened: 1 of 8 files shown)


justdan96:
Excellent! It feels like we're close to a point release and this is such a huge inclusion!

jcdr428:
@justdan96 @xavery yes , I think we're ready for 2.6.16 !!  👍

PR #113, opened 2020-01-06
FPS info read from VPS and SPS

jcdr428:
fps info can be either in VPS or SPS nal, or both, or none.
tsMuxer reads fps from VPS nal only.
This patch allows tsMuxer to find fps in SPS when not present in VPS.
Plus bug sorted out in flags reading (8 bits in lieu of 1 bit)...

### DIFF
--- a/tsMuxer/hevc.cpp
+++ b/tsMuxer/hevc.cpp
@@ -237,10 +237,7 @@ void HevcVpsUnit::setFPS(double fps)
 
 double HevcVpsUnit::getFPS() const
 {
-    if (num_units_in_tick != 0)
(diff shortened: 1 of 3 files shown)


justdan96:
Great stuff!

jcdr428:
Unfortunately I don't understand how the GUI extracts the fps from the streams, and I can't manage to have Qt5 working with MSVC2019. So I can't work on the GUI on my side...

justdan96:
Have you tried using the Docker container? It has Qt5 setup for Linux, MacOS and cross compiling for Windows.

jcdr428:
I got used to Visual Studio, but I'll have a look.

PR #122, opened 2020-01-08
Change 'famaly' to 'family' in the tsMuxerGUI settings

justdan96:
Please note this will mean current subtitle font family settings saved in your configuration would be lost.


xavery:
I'm all for correcting spelling mistakes - but we can easily keep backwards compatibility here. How about, if `family` doesn't exist, trying to read `famaly` in order to import the old setting?

justdan96:
You're right - I was breaking my own rule here "don't break current functionality". I'll add the functionality to check for "famaly" too.

jcdr428:
@justdan96 I hope you don't mind me jumping on your wip, but you seem busy -and I really would like to see version 2.6.16 released asap ;)
Replace
```
  QString fontName = settings->value("famaly").toString();
```
with
```
  // keep backward compatibility with versions < 2.6.15 which contain "famaly" key
  if (settings->contains("famaly")) {
    settings->setValue("family", settings->value("famaly"));
    settings->remove("famaly");
  }
  QString fontName = settings->value("family").toString();
```

justdan96:
Thanks for sorting this, I've been a bit busy recently!

xavery:
No worries - figured as much. :-)

PR #123, opened 2020-01-08
Warning on video format

jcdr428:
tsMuxer accepts V2 formats for V3 Blu-rays. A warning is shown when the format is not V3 compliant.

Cf. [https://forum.doom9.org/showthread.php?p=1895094#post1895094](https://forum.doom9.org/showthread.php?p=1895094#post1895094)


justdan96:
Seems reasonable enough to me!

PR #101, opened 2019-12-31
Correct Arrival Time Stamps

jcdr428:
The BD Player cannot read faster than the TS_Recording_Rate.
The TS_Recording_Rate is hereby fixed to be 6 MB/s for 2K (25/50 GB disk), and 13.625 MB/s for 4K (66/100 GB Disk).

Therefore the difference of ATS between two consecutive M2TS packets cannot be less than 846 PCR units for 2K, and 373 PCR units for 4K.
Those two values have been checked to be the standard for ATS difference between two video packets in several commercial HD and UHD Blu-rays.

This patch fixes BD-ROM Part3 Verifier Error M2TS0077 :

    ----------------------------------------------
    Error ID      : M2TS0077
    Target File   : 00000.m2ts <Dolby Digital audio PID=0x1100>
    Target Field  : <Source de-packetizer>.TS_recoding_rate
    Section No    : 6.2.2.1
    Error Message : TS_recoding_rate value is less than 188[byte] / (min(arrival_time_stamp[i] - arrival_time_stamp[i+1]) 
                / 27[MHz]). 
                
                TS_recoding_rate = 13625000.
                arrival_time_stamp[83073] = 177219240.
                arrival_time_stamp[83074] = 48874258294762232.
    Explanation   : The TS_recording_rate shall satisfy the following inequality. The TS_recording_rate 
                is expressed in units of bytes/second.
    - - - - - - - - - - - - - - - - - - - - - - -

Issue #108, opened 2020-01-04
Incorrect TS buffers management


jcdr428:
For the record:
TSMuxer calculates the ATS (Arrival Time Stamps) of each M2TS packet by simply dividing the time between two consecutive frames/DTS by the number of packets. This is incorrect (i.e. not complying with T-REC H.222.0 = ISO/IEC 13181-1) and creates constant underflow / overflow of the various TS buffers.

justdan96:
How does this affect #101 that was just merged?

jcdr428:
@justdan96 welcome back, happy new year !
#101 has only prevented the packets to be muxed with packets arriving faster than the transfer rate allows. It does not solve the way tsMuxer calculates a simplistic time gap between packets, by dividing the time between two video frames (1/fps) by the number of packets.
the way it should be done is quite well explained by drmpeg e.g. [here](https://forum.doom9.org/showthread.php?p=1105850#post1105850) and [here](https://forum.doom9.org/showthread.php?p=1576718#post1576718).

xavery:
I don't claim to be an expert when it comes to transport streams or the code that tsMuxer currently uses for packetisation, but is as far as I remember from my days in dabbling with MPEG-TS, it is always best to try and keep the average bitrate of the whole stream as constant as possible. This means that doing this properly is not possible without scanning every elementary stream in order to identify places where packets need to be more condensed, while at the same time keeping the PCR gap smaller than 100ms.

The posts you've linked provide a really good explanation. The official TS specification (the previous version is [freely available](https://www.itu.int/rec/T-REC-H.222.0-201703-S/en)) also provides some information in section 2.4 and Annex Q, but I have to admit that the language is hard to get through for somebody inexperienced. I can't see any viable solution in terms of code right now, especially not knowing the current state of things.

justdan96:
Are there any reference implementations we could crib off? Or do you think isolating the current implementation would be the first priority?

xavery:
[...]

PR #125, opened 2020-01-08
Revert on change of ATS #101

jcdr428:
Limiting the reading pace of packets to the Bluray Tranfer Rate value actually creates another issue -timegap between two PCR packets created by tsMuxer can be larger than 100ms.
This issue can be worse than the one solved...

I haven't managed to solve this new issue, so for the time being I prefer to revert on the previous merge #101.


justdan96:
What do you see the long-term fix for this being?

xavery:
I guess this can be tracked further as a part of #108. See my post in there.

Issue #89, opened 2019-12-27
Muxing m2ts with SRT file error


damkalan-uth:
When adding srt track on .m2ts file (HEVC 2160p HDR, AC3) it produces the following error.
Using Windows 10 

When removing  the Srt from input files it is ok.

![image](https://user-images.githubusercontent.com/59288630/71522828-08406280-28cf-11ea-9539-90935bfe0392.png)

justdan96:
If you save the meta file and run it through the tsMuxer CLI do you get the same error?

damkalan-uth:
D:\tsMuxer>tsMuxeR.exe 00004_2.meta
tsMuxeR version 2.6.15. github.com/justdan96/tsMuxer
Can't detect stream type

Directory of D:\tsMuxer

28/12/2019  02:01 πμ    <DIR>          .
28/12/2019  02:01 πμ    <DIR>          ..
27/12/2019  06:00 μμ       143.259.648 00004.m2ts
27/12/2019  05:20 μμ             1.380 00004.srt
28/12/2019  02:00 πμ               383 00004_2.meta
27/12/2019  01:08 πμ         3.739.546 tsMuxeR.exe
27/12/2019  01:09 πμ        21.445.632 tsMuxerGUI.exe
               5 File(s)    168.446.589 bytes

xavery:
I presume that the error appears because the subtitle line in the meta file contains `fps=0`, which is rejected by the program while attempting to mux.
The reason why `fps=0` appears is *probably* because tsMuxer fails to properly detect the frame rate while running it in single-file mode in order to detect the stream properties : however, I have not yet looked into this in more detail.

damkalan-uth:
Well neither CLI works. I forgot i removed SUBs ftom meta file. I ll try with fps edited

damkalan-uth:
Manually edited fps value to 23.976 and it worked!

xavery:
Okay, this seems to confirm that the issue is probably with how the frame rate is detected while adding streams to the GUI.

damkalan-uth:
To help, on video track Frame rate is detected correctly.

xavery:
Cannot reproduce with the current master. Could you try the [latest nightly build](https://bintray.com/justdan96/tsMuxer/tsMuxerGUI-Nightly/2019-12-27--01-10-05#files) and report if the issue still appears?

damkalan-uth:
No it didn't work with  latest nightly build.
However i did the following trick.
[...]

PR #130, opened 2020-01-10
Fix fps info

jcdr428:
Hopefully last bug on fps reading...

- correct text descriptions from sps and vps
- remove junk from metaDemuxer.cpp (calling getTrackFps which always sets fps to 0 !!)

### DIFF
--- a/tsMuxer/hevc.cpp
+++ b/tsMuxer/hevc.cpp
@@ -776,7 +776,7 @@ string HevcSpsUnit::getDescription() const
     result += (interlaced_source_flag ? string("i") : string("p"));
 
     double fps = getFPS();
-    result += " Frame rate: ";
+    result += "  Frame rate: ";
     result += (fps ? doubleToStr(fps) : string("not found"));
     return result;
 }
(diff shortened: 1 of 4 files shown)


jcdr428:
Should fix #89.

justdan96:
Has this been tested with a variety of sources, so we can check for regressions?

jcdr428:
@justdan96 I've tested with five files: .mkv, .ts, .hevc, hevc .m2ts, avc .m2ts
fps was fixed on these five tests.

xavery:
Wanted to merge this tonight, but I can't reproduce the issue no matter how hard I try. Let's wait until we get a reply on #89.

jcdr428:
@xavery agreed. The two samples I was sent (from W10x64  users, same as me) didn't produce any issue, it is strange that the behavior is not consistent between PCs.
What is sure is that the part with "getTrackFps" may have been added by Roman between 2.6.12 and 2.6.15 but seems unfinished.
Muxing AV and srt into an mkv is the only way I found to produce the issue. So let's see what is the feedbak...

jcdr428:
@xavery see my comment on issue #89: the bug comes from the decimal "comma" in the locale.
This fix #130 is still needed for hevc to work correctly.

xavery:
GUI issue reproduced and fixed. Roman was trying too hard to do the right thing and it came back to bite us on the butt - in his credit, though, locale-dependent behaviour in text parsing functions changed a bit between Qt 4 and Qt 5. I'll prepare a bigger commit which hopefully cleans this up. For now, this solves the immediate issue.

PR #131, opened 2020-01-10
Introduce a uniform formatting style

xavery:
This PR is here mainly for testing the actions subsystem when it comes to running clang-format.

Expect a lot of force pushes.


xavery:
Yay - works! I have purposefully not yet modified all the files "en masse", as I'm going to do that before actually merging this - like I said before, I'd like to do that once there are no outstanding PRs in order not to mess up anyone's work.

@jcdr428 @justdan96 I have incorporated the changes we discussed into this branch's clang-format file : post here if you'd like something else changed.

justdan96:
This is great!

jcdr428:
@xavery how will the line limit to 120 chars work ?
For instance ac3StreaMreader.cpp line 207:
`                m_lastMplsTime += (int64_t) ((m_mplsInfo[m_curMplsIndex].OUT_time - m_mplsInfo[m_curMplsIndex].IN_time)*(1000000000/45000.0));
`
where does the script decide to break the line ?

xavery:
> @xavery how will the line limit to 120 chars work ?
> For instance ac3StreamReader.cpp line 207:
> ` m_lastMplsTime += (int64_t) ((m_mplsInfo[m_curMplsIndex].OUT_time - m_mplsInfo[m_curMplsIndex].IN_time)*(1000000000/45000.0));`
> where does the script decide to break the line ?

This particular part is inside a `#if 0` and is thus ignored by clang-format, but the breaking usually occurs around operators. I deleted the `#if 0` and ran the formatter locally, this is what came out :
```
                m_lastMplsTime += (int64_t)((m_mplsInfo[m_curMplsIndex].OUT_time - m_mplsInfo[m_curMplsIndex].IN_time) *
                                            (1000000000 / 45000.0));
```

jcdr428:
@xavery any reason for the additional indents added on the second line to be there ?
i.e. why not have
```
                m_lastMplsTime += (int64_t)((m_mplsInfo[m_curMplsIndex].OUT_time - m_mplsInfo[m_curMplsIndex].IN_time) *
                (1000000000 / 45000.0));
```

xavery:
@jcdr428 I've never had to think about it that much. :-) Expression continuations are indented in order to correspond to the last term of the expression before they were broken. I don't think there's an option to disable that - at least looking for "expression", "continuation", or "indent" in the list of clang-format options doesn't produce anything that might be applied here. The only thing I tried setting was `AlignOperands: false`, but that had no effect.

jcdr428:
Ok, thanks @xavery. Actually the additional indent is because of the open bracket (BAS_Align), this makes sense after all.

xavery:
@jcdr428 @justdan96 Our source files have a mixture of CRLF and LF line endings - I'd like to standardise those too. Question is : which one to use? I don't think editors have problems with either these days, so this is really a matter of preference / main development platform that's used.

jcdr428:
@xavery LF not a problem for me, I work with Notepad++ and VS which keep both formats. And normally when I push with Git, it automatically changes CRLF to LF.

justdan96:
Agreed, keeping to /n makes things easier!

xavery:
LF it will be, then.

xavery:
It looks like major versions of clang-format might introduce changes to how the code is formatted - my local clang-format 9 produced different results than version 5 that's set to run on the GitHub runner.
@justdan96 The LLVM project [provides pre-built binaries](http://releases.llvm.org/download.html) for a lot of platforms - in this case, I see no real advantage in running the older version. If a contributor's system doesn't provide the latest version, the appropriate binary can be downloaded from the project site. Are you fine with running the latest version in this case?
@justdan96 @jcdr428 If you don't have any large local changes, I'll go ahead with merging this.

PR #139, opened 2020-01-15
Enhancement: Automatic V3 before muxing

jcdr428:
This patch allow automatic selection of BD-ROM V3 option in GUI before muxing when HEVC is detected.


jcdr428:
It will avoid this kind of misunderstanding https://forum.doom9.org/showthread.php?p=1895776#post1895776

PR #140, opened 2020-01-17
Fix 3d-plane / PID bug

### LINKED ISSUE #128: Bug: tsMuxer does not order 3D-Planes as per original MPLS
Cf. https://forum.doom9.org/showthread.php?p=1881372#post1881372  and https://forum.doom9.org/showthread.php?p=1882670#post1882670 :

"The other important bug with the subtitles is also related to the 3D BD. The MPLS file contains the extensions with the 3D-specific information, like what tsMuxeR calls the "3D-Planes", in fact the "Offset Sequences" necessary to display the subtitles at the correct depth on screen. There is usually one 3D-Plane per subtitle stream, although it's not mandatory as the same 3D-Plane can be used for several streams. The bug is that tsMuxeR assumes that the 3D-Planes are assigned to the subtitles streams in the order of their UID. It's not correct. The MPLS file describes precisely the subtitle streams, ans they do not have to be in the order of their UID, although it's often the case.

So, for example, if there are 4 subtitle streams in the BD, stored in the MPLS in the standard order, tsMuxeR will assign the 4 first 3D-Planes to that 4 streams, and it will probably be correct. But if the 4 streams are presented in the MPLS in another order, it will be wrong. And if the same 3D-Plane is assigned to several subtitle streams in the MPLS, it will assign it only once, and the last subtitle streams will not have any 3D-Plane at all. To summarize, tsMuxeR ignores the content of the MPLS when it assigns the 3D-Planes to the subtitle streams. That's an important bug, although the wrong information is only displayed, and never used during the demux operation. I don't know if tsMuxeR does the same error when it remuxes a 3D-BD, but I suppose so. "

jcdr428:
Cf. https://forum.doom9.org/showthread.php?p=1881372#post1881372 and https://forum.doom9.org/showthread.php?p=1882670#post1882670

PGS stream orders in mpls and clpi can be different.
PGS stream order for 3d-plane is taken from clpi , whereas it should be taken from mpls.

This patch fixes #128

### DIFF
--- a/tsMuxer/main.cpp
+++ b/tsMuxer/main.cpp
@@ -164,10 +164,11 @@ void detectStreamReader(const char* fileName, MPLSParser* mplsParser, bool isSub
 
             LTRACE(LT_INFO, 2, "Stream ID:   " << streams[i].codecInfo.programName);
             std::string descr = streams[i].streamDescr;
-            if (streams[i].codecInfo.codecID == CODEC_S_PGS)
+            if (streams[i].codecInfo.codecID == CODEC_S_PGS && mplsParser)
             {
                 // PG stream
-                int pgTrackNum = streams[i].trackID - 0x1200;
+                MPLSStreamInfo streamInfo = mplsParser->getStreamByPID(streams[i].trackID);
+                int pgTrackNum = streamInfo.streamPID - 0x1200;
                 if (pgTrackNum >= 0 && pgTrackNum < pgStreams3D.size())
                 {
                     if (pgStreams3D[pgTrackNum].offsetId != 0xff)


justdan96:
And here I was thinking this would be a huge amount of work to fix!

jcdr428:
@justdan96 well, yes it was a huge amount of work to fix ! It took me hours to find the origin of this bug, and as I didn't have any 3D bluray I had to recreate the bug by tweaking some BDs and inverting the MPLS streams with BDEdit.

As most of the bugs in tsMuxer, solution is trivial once identified ;)

PR #144, opened 2020-01-17
Use the checked out Git revision as app version if building in a Git repository

### LINKED ISSUE #127: Enhancement: add hash of tsMuxer latest commit after version number in console
It would allow to be certain of which version users have used when reporting bugs. Currently it is kind of a mess...

@xavery, I'm sure this is something you know how to implement ?

xavery:
This replaces the occurrences of the static version number with a version string
if building in a Git repository and Git is available on the build host.

This isn't perfect, since the revision is only going to be fetched when running
qmake or cmake, while it's possible that it changes between consecutive
compilations, not just reconfigurations. However, the main purpose of this is
to add a version tag to the builds built centrally by the GitHub runners, which
always configure the project from scratch.

Fixes #127.


justdan96:
Looks good to me!

xavery:
So let's roll.

PR #154, opened 2020-01-20
Remove all custom floating-point conversion functions

xavery:
QString::toDouble (and ::toFloat) behaviour changed between Qt4 and Qt5 : in
the older version, the function tried to convert the string according to the
current locale and, if unsuccessful, fell back to parsing according to the "C"
locale.

In Qt5, the conversion is always performed in the "C" locale, which means that
all the shenanigans related to finding and replacing the decimal point are not
needed anymore, since the main tsMuxer binary always runs in the "C" locale due
to not calling setlocale() at all.

Also, if Qt 4.8 documentation is to be believed, all this code wasn't ever
needed in the first place, as the conversion function would've just used the
"C" locale after failing to convert the string due to a mismatch in the used
decimal point character - and thus succeeded.


justdan96:
Great to see we are removing some more of the custom cruft in the code base!

PR #157, opened 2020-01-20
Fix bug 3d-plane / PID part 2

jcdr428:
All `pgStreams3D[pgTrackNum]` (coming from clpi) must be changed to `streamInfo` (coming from mpls). Tested with "The Hobbit : Desolation of Smaug" as per bug reported [here](https://forum.doom9.org/showthread.php?p=1896255#post1896255).

Also when 3d-plane is "FF", this means it is 'undefined' (i.e. 2D), not 'zero'.
Edit: added `(PID is not in mpls)` in decription in case of "phantom stream", i.e. PID present in clpi but not in mpls.

### DIFF
--- a/tsMuxer/main.cpp
+++ b/tsMuxer/main.cpp
@@ -169,23 +169,23 @@ void detectStreamReader(const char* fileName, MPLSParser* mplsParser, bool isSub
                 int pgTrackNum = streamInfo.streamPID - 0x1200;
                 if (pgTrackNum >= 0 && pgTrackNum < pgStreams3D.size())
                 {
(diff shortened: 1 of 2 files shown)


jcdr428:
@xavery why does clang, when I propose a patch on main.cpp, find me some corrections to be done in tsmuxerwindow.cpp ?

xavery:
> @xavery why does clang, when I propose a patch on main.cpp, find me some corrections to be done in tsmuxerwindow.cpp ?

Weird. I must've somehow missed this line when working on #154. Anyhow - thanks!

PR #160, opened 2020-01-22
Detect audio delay in MKV

jcdr428:
tsMuxer currently has the ability to detect audio delay from TS/M2TS/MPG/VOB/EVO sources, but not from MKV sources.

This patch adds ability to detect audio delay from MKV sources.


jcdr428:
Fixes issue #160.

PR #161, opened 2020-01-24
Implement support for default audio and subtitle tracks when muxing to Blu-ray

xavery:
This is a work in progress, as I would like to expose the functionality in the GUI as well.

The number of changes is large mostly because I implemented classes for creating the movie object data at runtime instead of having one array with the data and changing individual bytes depending on muxing parameters.


jcdr428:
@xavery note that there is already the function `int MovieObject::compose(uint8_t* buffer, int len, DiskType dt)` in tsPacket.cpp, although unused.

xavery:
F*ck me...

jcdr428:
@xavery yes tsMuxer is "a little" convoluted :) I added the note that the function was there but unused in #70 ...

justdan96:
@xavery I sense you are a little annoyed by that! I was hoping the code coverage reports would give us a nice list of unused functions but the format is a bit tricky for me to understand.

justdan96:
I like how this is shaping up - would you be able to update the summary with all the changes once you are done?

xavery:
Sure. I'm hoping to Get This Done™ during the weekend, hopefully I don't get too discouraged by the GUI, as I've already made two attempts at it and gotten frustrated halfway through.

xavery:
I think I'm done. The code in the GUI part is really shitty, but this is a natural consequence of the central `trackLV` widget being both a model and a view, so new code must be written around it to manage adding/removing items from other views like the new combo boxes for choosing the default track.
Let me know if you have any comments or questions, or if you think this should be merged.

xavery:
And the promised list of changes...
[*] Audio tracks whose parameter list in the meta file contains the `default` token will be marked as default tracks when muxing to Blu-ray.
[*] Subtitle tracks can also contain the `default` parameter with the value `all` - in which case all subtitles from the stream will be shown - or `forced`, which causes only subtitles marked as "forced" to be shown.
[*] The code for creating MovieObject data has been rewritten from scratch, and the old one in tsPacket.cpp has been removed. The infrastructure makes it easier to implement support for custom commands which can be added to the stream of MovieObjects inside the movie object file. The new code has been verified to produce the same results as the old one.
[*] The usage string in `main.cpp` now uses the [C++11 "raw string literals"](https://en.cppreference.com/w/cpp/language/string_literal) feature in order to throw out escaping, which made it cumbersome to keep the individual lines at 80 characters.
[*] Unused variables removed in `metaDemuxer.cpp` and `muxerManager.cpp`.
[*] Relevant functionality has been exposed in the GUI in the "Blu-ray" tab. The new options have an effect on the generated metafile only if muxing to Blu-ray.

justdan96:
Error from trying to compile for Win64:
```
../tsMuxerGUI/tsmuxerwindow.cpp: In lambda function:
../tsMuxerGUI/tsmuxerwindow.cpp:736:9: internal compiler error: Segmentation fault
         updateTracksComboBox(targetComboBox);
         ^
x86_64-w64-mingw32.static-g++: internal compiler error: Segmentation fault (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
```

xavery:
Compiler segfault with valid code ... now that's a first for me. 🤣

justdan96:
More likely to be a bug in the version of GCC MXE is using, I can see if there is a newer version available?

xavery:
Yes, most probably. I suspected this might be related to a lambda being used here, and it turns out I was right, so I rearranged the code a bit in order to get rid of it. Pushed the fix to master directly.

PR #163, opened 2020-01-24
Remove 3d-plane info for 2D BD-ROM

jcdr428:
Fix  bug provoked by #159: 3d-plane information should show only for 3D BD-ROMs


justdan96:
Seems like a straightforward bug fix

PR #169, opened 2020-01-31
Bug: Shorten filename when adding more than two files

jcdr428:
See eg https://forum.doom9.org/showthread.php?p=1897783#post1897783

tsMuxer currently adds the filenames one after the other to create the new file name.
This creates a "Can't create output file" error on Windows when new file name is more than 255 bytes.

This patch simply names the created files "first file+...+lastfile.ext" when more than two files are concatenated.

### DIFF
--- a/tsMuxer/singleFileMuxer.cpp
+++ b/tsMuxer/singleFileMuxer.cpp
@@ -121,7 +121,7 @@ void SingleFileMuxer::intAddStream(const std::string& streamName, const std::str
 
     vector<string> fileList = extractFileList(streamName);
     string fileName;
-    if (fileList.size() < 2)
+    if (fileList.size() < 3)
         for (int i = 0; i < (int)fileList.size(); i++)
         {
             if (i > 0)
--- a/tsMuxer/singleFileMuxer.cpp
+++ b/tsMuxer/singleFileMuxer.cpp
@@ -129,7 +129,7 @@ void SingleFileMuxer::intAddStream(const std::string& streamName, const std::str
             fileName += extractFileName(fileList[i]);
         }
     else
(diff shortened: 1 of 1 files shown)


abakum:
>This creates a "Can't create output file" error on Windows
On linux too
>when more than two files are concatenated.
More than one
>"first file+...+lastfile"
Maybe better "first file+_+lastfile"? 
Or "first file_lastfile" so that the scripts are not confused with .ext

Thanks for the quick help!

abakum:
forfiles /M *track_* /C "cmd /c del track.@ext&ren @file track.@ext"
Worked fine

justdan96:
Is it the filename length that is limited at 255 or the overall path length?

jcdr428:
@justdan96 The Windows default path length limitation (MAX_PATH) is 256 characters.

abakum:
https://stackoverflow.com/questions/265769/maximum-filename-length-in-ntfs-windows-xp-and-windows-vista

justdan96:
Maybe we could add an error if the file/path length exceeds those values? We could also add a check on acceptable file names if we wanted to be thorough.

jcdr428:
@justdan96 you can have many cases where the resulting filename will be above 255 chars with current tsMuxer: eg the movie in a blu-ray can be split into >30 m2ts (common with eg Disney prods) and you want to concatenate all of them together.

abakum:
"Best the enemy of the good"
Thanks, everything works so well!

justdan96:
The "..." makes me a little nervous, maybe an underscore would be more appropriate? I'm just thinking of programs that detect the extension based on the first dot character found.

abakum:
first file-lastfile

jcdr428:
@justdan96 no worries... changed by underscores___

justdan96:
Cool, looks good!

PR #171, opened 2020-02-02
Bug: Subtitle PIDs must start at 0x12A0 for HDR

jcdr428:
tsMuxer currently gives subtitle PIDs the value of 0x1200 and above.

For BD V3 m2ts with HDR, PGS PID must start at 0x12A0 -thanks @tymoxa for raising the issue.


jcdr428:
@xavery something wrong with the clang check, started 20 min. ago and still in progress ?

xavery:
@jcdr428 Seems like a hiccup at GitHub. No worries.

PR #173, opened 2020-02-03
Bug: "Secondary" box greyed out for DTS Express 24bit

jcdr428:
The checkbox is accessible for DTS Express or DD+, but not for DTS Express 24-bit.

This simple patch fixes this.


justdan96:
Great stuff!

PR #174, opened 2020-02-04
For MSYS2

abakum:
Instead of mingw-w64-i686 for 32bit or mingw-w64-x86_64 for 64bit, it’s convenient to use $MINGW_PACKAGE_PREFIX
Before use cmake ../ -G Ninja need install ninja.
Instead of cmake . --build use cmake --build . or use rebuild_linux.sh

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -208,32 +208,18 @@ flex \
 zlib-devel
 ```
 
-32-bit:
 ```
-pacman -Sy --needed mingw-w64-i686-toolchain \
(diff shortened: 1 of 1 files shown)


justdan96:
That's a great idea! I'll give @xavery a chance to give his opinion too but thanks for the contribution!

abakum:
image

xavery:
LGTM. I'm all for reducing duplicated instructions if they can be replaced with a variable reference.

PR #180, opened 2020-02-06
Now the tsMuxeR requires .meta in utf8

abakum:
Thanks, @xavery  for unicode


justdan96:
What exactly was the format that the meta file was in before? I thought it was ASCII

xavery:
The encoding of the meta file was actually dependent on the code page of the computer where the program was running - talking about Windows, of course. Both Linux and MacOS have always used UTF-8 in this regard.

justdan96:
Ah of course. Maybe we can tighten up the wording to say it only affects Windows?

xavery:
Yup. I'll push some wording fixes and merge this later on.

abakum:
Great, then I close this request

PR #181, opened 2020-02-06
Try using the active code page if conversion from UTF-8 fails

xavery:
This makes the program compatible with meta files saved in ACP by older GUI
versions.


xavery:
Since having false positives is possible (but very unlikely) by using this backwards compatibility shim, I also put an appropriate note in the README.

abakum:
@xavery please add ACP-UTF detection for *.srt files. Right now tsmuxer only works correctly with *.srt in UTF-8 with BOM. It would be great if it worked with them in ACP encoding or in UTF-8 without BOM and in UTF-8 with BOM for compatibility.

abakum:
Like "Try using the active code page if conversion from UTF-8 fails"

xavery:
I was certain this was already implemented this way ... will have a look later on.

PR #182, opened 2020-02-07
Use rebuild_msys2.sh

abakum:
Thanks @xavery for qmake guide


abakum:
https://github.com/justdan96/tsMuxer/pull/183

PR #183, opened 2020-02-07
MSYS2 build script

abakum:
Thanks @xavery for qmake guide

### DIFF
--- a/rebuild_msys2.sh
+++ b/rebuild_msys2.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/bash
+
+pacman -Syu
+if [ "$MSYSTEM" == "MSYS" ] ; then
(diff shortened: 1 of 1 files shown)


abakum:
https://github.com/justdan96/tsMuxer/pull/182

justdan96:
I think we should separate out the build requirements from the build process - otherwise we risk trying to install the same things each time we rebuild. Also can you please merge all changes into 1 pull request please?

abakum:
You're right but for non-sophisticated Windows users, it would be convenient to just run one script.
The --needed pacman's option will not allowed to install the build environment twice.
Checking for the presence of the build directory if [! -d build] will not allowed to run the installation of dependent packages every time.
I will try to learn how to collect several changes in one request.

PR #186, opened 2020-02-08
Detect UTF-8 in SRT files

xavery:
Also, remove unused code in utf8Converter.cpp and remove the "sfDefault" from
the SourceFormat enum, whose meaning was platform-dependent anyway and did not
make sense anymore, as we now default to UTF-8 on non-Windows platforms anyway.
Processing falls back to the ACP on Windows to keep backwards compatibility.


justdan96:
Great work!

PR #207, opened 2020-02-25
Fix reading MP4 (bug from #205)

jcdr428:
With patch #205, tsMuxer is reading over the end of the mp4 file, which results in erratic behavior -Visual Studio build works correctly, but cmake build does not.

This fix allows to stop the reading of the mp4 file at the last atom, before end of file.


jcdr428:
@justdan96 any problem merging this fix ?

justdan96:
Looks good to me

jcdr428:
@justdan96 the tmpFile is just used to open the meta file, get its byte size, and close it.

PR #213, opened 2020-02-28
[bug]Fix subtitle timestamp

jcdr428:
When joining several m2ts together (either in a mpls or manually), tsMuxer takes into consideration the start-time of the first file only.
This results in an incorrect subtitle start time when the m2ts don't all have the same start-time.

The bug has been reported e.g. [here](https://forum.doom9.org/showthread.php?p=1768965#post1768965) and [here](https://forum.doom9.org/showthread.php?p=1881372#post1881372).

This simple fix solves this bug.


justdan96:
Looks straightforward to me

PR #215, opened 2020-02-28
Support of m4v files

jcdr428:
m4v files have the same structure as mp4 -only change is the extension filename.

Tested with https://standaloneinstaller.com/blog/big-list-of-sample-videos-for-testers-124.html

What format is m4v: https://en.wikipedia.org/wiki/M4V


justdan96:
As soon as I saw M4V I got flashbacks from the painful migration away from iTunes I did years ago :D

Looks like a straightforward change to me.

PR #216, opened 2020-02-28
[bug]Incorrect subtitle space in Windows

jcdr428:
The font space between regular and italics is incorrect, simply because the pen object was not passed to the GDI+ 'GetBound' class !

The bug was reported e.g. [here](https://forum.doom9.org/showthread.php?p=1752780#post1752780) and [here](https://forum.doom9.org/showthread.php?p=1898949#post1898949).


jcdr428:
Grrr... four hours to find the solution to this bug !

PR #219, opened 2020-03-01
Remove all usages of std::wstring

xavery:
I've never been a fan of `std::wstring`, seeing how it does not seem to serve any actual purpose except giving people the false sense of being cross-platform when using it.
Unfortunately, and as noted in the comments in `convertUTF.h`, using `wchar_t` (on which `std::wstring` is based on) is no less portable than using `unsigned short`. I wholeheartedly agree, which is why this commit erases all usages of `std::wstring` from the face of the project and reduces `wchar_t` usages to the necessary minimum, mostly for interoperability with WinAPI functions. UTF-8 encoded `std::string`s are used in place of `std::wstring` objects, and code which must operate on actual characters instead of the bytes which might represent a character has been modified in order to work properly - mostly the subtitle renderer and reader classes.
Short summary of other changes :

- Stuff from `convertUTF.h` now lives in the `convertUTF` namespace.
- Needlessly duplicated `*W` variants of custom string functions have been removed.
- `defaultPalette` in `TextSubtitlesRender` is now an `unordered_map` instead of an array using linear lookup.
- `utf8Converter.cpp` has been practically rewritten from scratch in order to get rid of the `if`s on `sizeof(wchar_t)`, which are now needless, since the only function that's left is used to convert characters to UTF-8 from other encodings.
- `SourceFormat::sfANSI` is now `#ifdef _WIN32`, since there is no point in pretending that this value has any meaning outside of Windows.


jcdr428:
@xavery 'Note that *legal* UTF-8 values can't have 4 or 5-bytes' : to my knowledge, in RFC 3629 UTF-8 values can have up to 4 bytes ?

xavery:
@jcdr428 That comment refers to the trailing bytes that can occur in a valid UTF-8 sequence. The algorithm allows 4 or 5 bytes (making the length of the entire sequence 5 or even 6 bytes) of trailing, but those sequences are deemed invalid due to the fact that a 4-byte sequence already contains 21 bits for the actual code point which is deemed enough to hold all possible Unicode values, as they will never be higher than U+10FFFF.

jcdr428:
Ok, thanks @xavery .

PR #222, opened 2020-03-01
Fix broken ISO labels when using non-ASCII characters

### LINKED ISSUE #185: MUXOPT option --label for .iso does not work with non ASCII characters
If iso.meta with
MUXOPT --blu-ray --label="20200111 ДР Аллы" --vbr --vbv-len=500
... 
encoded in windows-1251 result after
tsMuxeR.exe iso.meta "20200111 ДР Аллы.iso"
like:

If iso.meta encoded in UTF-8 result like:

It would be great to have it like after imgBurn:

xavery:
This commit fixes the broken ISO labels when using characters outside of the character set supported by ISO-8859-1. Every dstring written to the UDF headers is now inspected whether it can use the limited encoding or if it's necessary to encode it as 16-bit. This has the advantage of leaving all the dstrings which don't need 16-bit encoding, like folder and file names, without any modification in the file structure.
Tested with the following metafile :
```
MUXOPT --no-pcr-on-video-pid --new-audio-pes --blu-ray --label="для Константина" --vbr  --auto-chapters=5 --vbv-len=500
V_MPEG4/ISO/AVC, "/home/daniel/tsMuxer_samples/video.h264", insertSEI, contSPS
```
Result :
Screenshot_2020-03-01_21-32-45

@abakum :wave: 
Unfortunately, this cannot be merged right away since the lambdas cause crashes when building in MXE. We've seen this before :
```
/workdir/tsMuxer/iso_writer.cpp: In instantiation of '{anonymous}::serializeDString(const string&, int)::<lambda(auto:3)> [with auto:3 = unsigned int]':
/workdir/tsMuxer/convertUTF.h:200:23:   required from 'void convertUTF::IterateUTF8Chars(const string&, Fn) [with Fn = {anonymous}::serializeDString(const string&, int)::<lambda(auto:3)>; std::__cxx11::string = std::__cxx11::basic_string<char>]'
/workdir/tsMuxer/iso_writer.cpp:146:10:   required from here
/workdir/tsMuxer/iso_writer.cpp:138:53: internal compiler error: in tsubst_copy, at cp/pt.c:13040
     const auto maxHeaderAndContentLength = fieldLen - 1;
                                                     ^
/workdir/tsMuxer/iso_writer.cpp:138:53: internal compiler error: Segmentation fault
i686-w64-mingw32.static-g++: internal compiler error: Segmentation fault (program cc1plus)
```
@justdan96 Since this code is entirely valid and this is undoubtedly a bug in the compiler itself, could you look into upgrading it?

Fixes #185 .

### DIFF
--- a/tsMuxer/convertUTF.cpp
+++ b/tsMuxer/convertUTF.cpp
@@ -103,6 +103,20 @@ ConversionResult ConvertUTF32toUTF16(const UTF32** sourceStart, const UTF32* sou
     return result;
 }
 
+std::tuple<UTF16, UTF16> ConvertUTF32toUTF16(UTF32 ch)
+{
+    if (ch <= UNI_MAX_BMP)
+    {
+        return std::make_tuple(static_cast<UTF16>(ch), 0);
+    }
+    else
+    {
+        ch -= halfBase;
+        return std::make_tuple((UTF16)((ch >> halfShift) + UNI_SUR_HIGH_START),
+                               (UTF16)((ch & halfMask) + UNI_SUR_LOW_START));
+    }
+}
+
 /* --------------------------------------------------------------------- */
 
 ConversionResult ConvertUTF16toUTF32(const UTF16** sourceStart, const UTF16* sourceEnd, UTF32** targetStart,
--- a/tsMuxer/convertUTF.h
+++ b/tsMuxer/convertUTF.h
@@ -89,6 +89,7 @@
 
 #include <cstdint>
 #include <string>
+#include <tuple>
 
 namespace convertUTF
 {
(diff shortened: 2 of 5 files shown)


abakum:
Via mingw is going great!
image
In tsMuxer/osdep/textSubtitlesRenderFT.cpp error

justdan96:
I can't find the exact bug in GCC 5.5, might have to check if Clang has the same issue. Apparently there is a way to use GCC 8 in MXE but I need to do a bit more research.

abakum:
> Via mingw is going great!

# gcc -dM -E -xc /dev/null | grep __GNUC__
#define __GNUC__ 9

justdan96:
Hopefully this should be working fully in the Docker container now. I'll do a test build before merging this change though.

xavery:
Yup, works great. Merging.

PR #226, opened 2020-03-03
Simplify TsMuxerWindow::myPlaySound

xavery:
It turns out that playing sound files available as application resources is as simple as using `QSoundEffect` with a `qrc` URL. This saves us from all the shenanigans previously done in order to save the resource to a file just to be able to pass it to a `QSound` object.


jcdr428:
Nice !

PR #227, opened 2020-03-03
Investigate building the GUI for Windows XP

xavery:
Yep, I'm really doing it.

### DIFF
--- a/.github/workflows/windows_xp_gui.yml
+++ b/.github/workflows/windows_xp_gui.yml
@@ -1,0 +1,19 @@
+name: Build GUI for Windows XP
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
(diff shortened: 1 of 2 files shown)


xavery:
I'd say this is ready. The only deficiency that's left is the lack of translations, which is due to QMake in Qt 5.6.3 not supporting the `lrelease` and `embed_translations` tokens in the `CONFIG` variable which our `.pro` file uses due to them being very convenient in our usecase.
@justdan96 The main motivation behind this was providing this executable inside the w32 package that's uploaded to bintray, since this way the whole package becomes runnable on Windows XP and very old CPUs. Let me know if you'd be willing to do that, if not - having this merged in would be good too, as it would check the compatibility with Qt 5.6.3 with every pull request that touches GUI stuff.

justdan96:
👍

PR #228, opened 2020-03-04
Сhanged address of test file

abakum:
from uhdsample.com to 4kmedia.org
image

image


justdan96:
Seems fair enough to me! I'll merge as soon as you give the okay.

PR #230, opened 2020-03-04
Fix Filename Truncation on Dot

### LINKED ISSUE #225: Filenames with periods get truncated on demux
If the input filename of a demux operation contains periods besides the last one then the filename gets truncated. For example "Trip to the U.S. West Coast.ts" becomes something like "Trip to the U.track_256.264" instead of "Trip to the U.S. West Coast.track_256.264", like any other file.

Since the filenames of the output of a demux operation can't be controlled then it's crucial that all files follow the same logic, in order to use it in scripts for automation..

justdan96:
method to extract the filename reads the string from right to left, this change ensures that when we extract the filename it will stop on the first period encountered

Fixes #225

### DIFF
--- a/libmediation/types/types.cpp
+++ b/libmediation/types/types.cpp
@@ -300,7 +300,7 @@ string extractFileName(const string& src)
     for (int i = src.size() - 1; i >= 0; i--)
         if (src[i] == '.')
         {
-            endPos = i;
+           if (endPos == src.size()) endPos = i;
         }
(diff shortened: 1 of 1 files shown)


jcdr428:
@justdan96 actually as the filename is read from right to left, it should be "ensure that when we extract the filename it **will** stop on the first period encountered" ;)

justdan96:
Thanks, will change the decription

PR #231, opened 2020-03-04
Always keep M2TS descriptors in TS files

jcdr428:
Regression on #203 Ts descriptors.
Until a patch is proposed to introduce an option to force M2TS descriptors in TS files.


justdan96:
This seems like a decent idea until we can get a long-term solution.

PR #238, opened 2020-03-09
cp tsMuxerGUI bin/

abakum:
On mingw32 tsMuxerGUI is going to build/debug/ and on mingw64 now in build/


justdan96:
Isn't there a way of specifying we want a release rather than a debug build?

PR #243, opened 2020-03-13
Non blu ray dolby vision

jcdr428:
In non-Bluray mode (TS or M2TS), when Dolby Vision is detected, include the 'DoVi' Registration Descriptor and the Dolby Vision video descriptor.

The descriptors are still HDMV, for compatibility with PGS, AC3/TrueHD and AC3/EAC3 streams.

Also includes:
 - DV descriptors moved from tsPacket to hevcStreamReader class;
- Correction of non HDMV AVC descriptor, for future use;
 - non HDMV HEVC descriptor, for future use;
 - declaration/enumeration of V3 flags, moved from hevcStreamReader to tsMuxer;

Note that there is a bug on Bluray players that I haven't managed to get rid of (see [discussion on doom9](https://forum.doom9.org/showthread.php?p=1903300#post1903300)), but I came to the conclusion that it doesn't come from the descriptors but rather from the way the packets are muxed. I will look at this separately.

### DIFF
--- a/tsMuxer/aacStreamReader.h
+++ b/tsMuxer/aacStreamReader.h
@@ -10,7 +10,7 @@ class AACStreamReader : public SimplePacketizerReader, public AACCodec
    public:
    public:
     AACStreamReader() : SimplePacketizerReader(){};
-    int getTSDescriptor(uint8_t* dstBuff, bool isM2ts) override { return 0; }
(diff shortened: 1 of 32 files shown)


jcdr428:
Note: @filip11111 has confirmed that these descriptors for Dolby Vision are working on the Sony x700 player -except for the pixelisation bug when stopping- see https://github.com/jcdr428/tsMuxer/pull/60#issuecomment-598955313

PR #248, opened 2020-03-18
Update hevcStreamReader.cpp

jcdr428:
Set hdr flags only if hevc stream (i.e. sps, pps, vps) has been detected.

This is a correction brought to #243 .


justdan96:
Great stuff, thanks!

PR #250, opened 2020-03-20
Correct DV Profile 4

jcdr428:
DV Profile 4 is single track double layer
DV Profile 7 is double track double layer

See https://forum.doom9.org/showthread.php?p=1904144#post1904144


justdan96:
Nice, so previous to this all Single Track Double Layer files were treated as Single Track Single Layer files and now we can properly handle those cases. Great work!

PR #254, opened 2020-03-22
For double video track, DV EL stream type = 0x06

jcdr428:
As per 3.1.3.1 of DolbyVisionProfilesLevels_v1_3_2_2019_09_16.pdf, for the secondary Dolby Vision PID carrying the EL and RPU substream, "the value of stream_type shall be set to 0x06 (indicating PES packets containing private data)."


justdan96:
Nice one!

PR #260, opened 2020-03-26
[bug] Fix num_applicable_ops_minus1 size

jcdr428:
Fixes bug notified here: https://forum.doom9.org/showthread.php?p=1904881#post1904881

Note that it fixes the "vector subscript out of range" error, it does not fix the fact that the notification window does not show any error message !


justdan96:
I won't pretend to understand exactly what is going on but as long as we are squashing bugs that's all that matters!

Issue #262, opened 2020-03-27
[bug] With multi-m2ts mpls, subtitles drift out of sync


jcdr428:
See notification of bug here https://forum.doom9.org/showthread.php?p=1905010#post1905010

FilipeAmadeuO:
The solution maybe would be for tsmuxer to mantain the multi m2ts (one mpls) struture and not join the multi m2ts in one big m2ts.
This way we could build one by one m2ts and keep the exact same struture as before.

jcdr428:
@FilipeAmadeuO that would be nice, however this would be a substantial rewrite of the code as currently tsMuxer is set to write only one m2ts, clpi, mpls, and pre-defined index.bdmv and MovieObject.bdmv. Plus code writing for the ATCd seamless branching etc.

FilipeAmadeuO:
> @FilipeAmadeuO that would be nice, however this would be a substantial rewrite of the code as currently tsMuxer is set to write only one m2ts, clpi, mpls, and pre-defined index.bdmv and MovieObject.bdmv. Plus code writing for the ATCd seamless branching etc.

OK. Understood. Maybe a improvement for the future.

lonecrane:
According to the tests on the UHD BD 'WALL E', now subtitles are demuxed very precise.
Internal timestamp presented by player:
![fig 1-2](https://user-images.githubusercontent.com/36885830/77837079-606d5600-7197-11ea-8de0-6887b5b6f917.png)
The Demuxed timestamp with the latest version matches very precisely:
![fig 1-1](https://user-images.githubusercontent.com/36885830/77837083-63684680-7197-11ea-96cf-d51c5acf82c7.png)
The Demuxed timestamp with the previous version w64-nightly-2020-03-25--01-12-53 delays 358ms:
![fig 1-3](https://user-images.githubusercontent.com/36885830/77837149-50a24180-7198-11ea-8612-0a66a8a61f2d.png)

PR #263, opened 2020-03-28
[bug] Fix Subtitle Delay when concatenating M2ts

jcdr428:
For subtitles, tsMuxer currently calculates the length of each m2ts relative to the last PTS.
The length of the m2ts should be calculated relative to the last video PTS, as there can be some audio/subtitle tracks longer than the video track.

Should fix issue #262 .


lonecrane:
Amazing! Can this out-sync mechanism explain the similar issue with regard to audio streams?

jcdr428:
@lonecrane I suspect something similar, I am looking at it. But the current fix solves the subtitle delays only.

PR #265, opened 2020-03-28
[bug] Add H265 to isVideoPID()

### LINKED ISSUE #258: Some demuxed audio streams are out of sync (maximum delay add up to -0.5s)
Hi,

I've noticed obvious 'out of sync' issue for some audio streams demuxed from a BD. This is the 'WALL·E' UHD blu-ray produced by Disney, and the playlist 00800.mpls consists of 42 m2ts segments 56,61,57 and so one.  The audio streams out of sync include DTS-HD Master streams and Dolby ac3 streams, but with the exception of Dolby TrueHD. At the very beginning the V/A are in sync, but they lose sync gradually. At the very end, the audio streams behave a maximum delay of about -0.5s.
I have made some tests and it appears that txmuxer improprialy changes some audio frames close to the beginning of the segment #57 (but not the first frames). 
fig 1
- The top track is a DTS-HD Master stream demuxed with tsmuxer from the BD and then converted to flac:
`eac3to 00800.track_4353.dts 00800.track_4353.flac -mono`
- The second track is the DTS stream from 00061.m2ts, which is the second segment of the film.
`eac3to BDMV\STREAM\00061.m2ts 3: 1_61_3.dts`
`eac3to 1_61_3.dts 1_61_3.flac -mono`
- The third track is the the DTS stream from 00057.m2ts, which is the third segment of the film.
`eac3to BDMV\STREAM\00057.m2ts 3: 2_57_3.dts`
`eac3to 2_57_3.dts 2_57_3.flac -mono`

I've aligned segments 61 and 57 so that the audio frames overlap. We can find that tsmuxer has correctly concatenate the two segments at the time about 0:1:04.100. However, at the time mark #1 (0:1:04.783) the top track begins to be different from the third track:
fig 2
But in the next, if we align the time mark #2 of first track with the time mark #3 of the third track, the subsequent waveforms seem identical again. The duration from time mark #1 to #2 for the first track is only 6ms, but the actual duration from time mark #1 to #3 for the third track is 17ms, thus a time delay of -11ms is introduced here.
The timeshift will be accumulated by the subseqent segments, so maybe that's the very reason that why the overall time delay reaches -0.5s.
I had found this issue for a long time, but I didn't have proper tools (here is audacity 2.3.3, a freeware) to demonstrate it until I find similar bug report in another forum.

jcdr428:
Necessary (but might not be enough) to fix #258 .

### DIFF
--- a/tsMuxer/tsDemuxer.cpp
+++ b/tsMuxer/tsDemuxer.cpp
@@ -165,7 +165,8 @@ bool TSDemuxer::isVideoPID(uint32_t streamType)
 {
     return streamType == STREAM_TYPE_VIDEO_MPEG1 || streamType == STREAM_TYPE_VIDEO_MPEG2 ||
            streamType == STREAM_TYPE_VIDEO_MPEG4 || streamType == STREAM_TYPE_VIDEO_H264 ||
-           streamType == STREAM_TYPE_VIDEO_MVC || streamType == STREAM_TYPE_VIDEO_VC1;
+           streamType == STREAM_TYPE_VIDEO_MVC || streamType == STREAM_TYPE_VIDEO_VC1 ||
+           streamType == STREAM_TYPE_VIDEO_H265;
 }
 
 // static uint64_t prevDts = 0;


justdan96:
Nice one!

Issue #261, opened 2020-03-26
[bug] Full path name is elided in source file column


jcdr428:
See https://forum.doom9.org/showthread.php?p=1904982#post1904982

jcdr428:
Fixed with #266 .
I suppose that QTableWidget WordWrap was off by default with qt4.

PR #266, opened 2020-03-28
Correct Eliding of track filenames

jcdr428:
Set WordWrap to false so that the eliding occurs at each character and not at each word.

Fixes issue #261 .


justdan96:
...and today I learned a new word😁

jcdr428:
@justdan96 same for me :)

PR #269, opened 2020-03-29
[bug] Integration of old Dolby Vision profiles 2, 3 and 6

jcdr428:
EDIT:
So it started with a small correction, but the more I search on DV profiles, the more I find...
From https://forum.blu-ray.com/showpost.php?p=12875562&postcount=645 I have now included DV profiles 2, 3 and 6 (deprecated on new equipment).

Plus As per T-REC-H.265 Annex E.3 VUI Semantics, Colour Primaries, Transfer Characteristics and Matrix Coefficients, when not present in the VUI, should be = 2 (Unspecified) by default. Value 0 is reserved for future use.

Plus code simplification: DVCompatibility is not compulsory in the DoVi descriptor.
Plus change HDMV descriptor to non-HDMV 'HEVC' registration descriptor for HEVC DV tracks.

### DIFF
--- a/tsMuxer/hevc.cpp
+++ b/tsMuxer/hevc.cpp
@@ -271,11 +271,9 @@ HevcSpsUnit::HevcSpsUnit()
       vcl_hrd_parameters_present_flag(false),
       sub_pic_hrd_params_present_flag(false),
       num_short_term_ref_pic_sets(0),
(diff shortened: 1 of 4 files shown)


jcdr428:
@justdan96 this commit is now complete and ready to be merged.

PR #274, opened 2020-04-07
Check for Invalid Characters in Output Filename

### LINKED ISSUE #268: Output path check for invalid characters
I was making TS Muxing, and have forgot about invalid characters in Windows file names. So i specified output folder with file name "Alita: Battle Angel (2019).ts". After muxing completed, a file with name Alita was created in output folder. It's size was 0 bytes. 
image
image
But the file itself is flushed to the disk and takes space. You can see it in file recovery software
image
I think the could be implemented check for invalid characters in output path before starting muxing.

justdan96:
We use different regex patterns for Unix and Windows to check for invalid characters in the output filename. If invalid characters are found we throw a runtime error.

Additionally some changes were made to make sure there was more uniform use of the getDirSeparator() function.

Should resolve #268 

Output strings tested:

```
Alita: Battle Angel (2019).ts
Alita Battle Angel (2019).ts
Alita Battle Angel (2019)для Константина.ts
```

Only the first should fail on Windows.

### DIFF
--- a/libmediation/types/types.cpp
+++ b/libmediation/types/types.cpp
@@ -25,11 +25,11 @@ using namespace std;
 regex invalidChars()
 {
 #ifndef _WIN32
-    // NULL and /
-    regex invalid("[\\/\\x00]");
+    // / and ASCII 0 to 31
+    regex invalid("[/\x00-\x1F]");
 #else
-    // <>:"/|?* and ASCII 0 to 31
-    regex invalid("[:<>I\"\\/|?\\*\\x00-\\x1F]");
+    // <>:"/|?\* and ASCII 0 to 31
+    regex invalid("[:<>\"/|?\\*\x00-\x1F]");
 #endif
     return invalid;
 }
--- a/libmediation/types/types.cpp
+++ b/libmediation/types/types.cpp
@@ -5,6 +5,7 @@
 #include <cstring>
 #include <iomanip>
 #include <ostream>
+#include <regex>
 #include <sstream>
 
 #ifndef _WIN32
@@ -21,15 +22,16 @@
 
 using namespace std;
 
-bool isInvalidChar(const char& ch) 
+regex invalidChars()
 {
 #ifndef _WIN32
-	// NULL and /
-    return (int(ch) == 0 || ch == getDirSeparator()); 
+    // NULL and /
+    regex invalid("[\\/\\x00]");
 #else
-	// <>:"/|?* and ASCII 0 to 31
-    return (ch == '<' || ch == '>' || ch == ':' || ch == '"' || ch == '/' || ch == '|' || ch == '?' || ch == '*' || ch == getDirSeparator() || int(ch) <= 31); 
+    // <>:"/|?* and ASCII 0 to 31
+    regex invalid("[:<>I\"\\/|?\\*\\x00-\\x1F]");
 #endif
+    return invalid;
 }
 
 uint64_t my_ntohll(const uint64_t& original)
@@ -367,16 +369,11 @@ string closeDirPath(const string& src, char delimiter)
 bool isValidFileName(const string& src)
 {
     string filename = extractFileName(src);
-    // isValidChar defined in file_win32.cpp and file_unix.cpp
-    string::iterator it = find_if(filename.begin(), filename.end(), isInvalidChar);
-    if (it != filename.end())
-    {
-        return false;
-    }
-    else
-    {
-        return true;
-    }
+
+    // invalidChars() returns a different regex pattern for Windows or Unix
+    regex ourInvalidChars = invalidChars();
+    bool isvalid = !(std::regex_search(filename, ourInvalidChars));
+    return isvalid;
 }
 
 string trimStr(const string& value)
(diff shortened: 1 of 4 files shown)


jcdr428:
@justdan96 `regex invalid("[:<>I\"\\/|?\\*\\x00-\\x1F]");`

Why do you hate the 'I' so much ?

justdan96:
It's supposed to be a pipe! I'll look at correcting that.

jcdr428:
@justdan96 do we really want unix tsMuxer to allow characters such as * or :, with the result that it will output some ts files that cannot be universally read by all OS ? i.e. why not imposing the `[:<>\"\\/|?\\*\\x00-\\x1F];` rule to all OS ?

justdan96:
I'm not sure if it is better to set limits on valid characters the same across the board - it could break some existing valid behaviour.

xavery:
> @justdan96 do we really want unix tsMuxer to allow characters such as `*`, `?`, `"` or `:`, with the result that it will output some ts files that cannot be universally read by all OS ? i.e. why not imposing the `[:<>\"\\/|?\\*\\x00-\\x1F]` rule to all OS ?

This is a decades-long debate about how the POSIX rule about allowing any pathname characters except nuls and forward slashes actually creates more confusion than usability, however the context in which I've seen it being doubted usually had something to do with script languages, though not necessarily. If you *really* want to get deep, deep down, have a look at [Fixing Unix/Linux/POSIX Filenames by David A. Wheeler](https://dwheeler.com/essays/fixing-unix-linux-filenames.html). The conclusion of this article is that limiting the set of allowed characters would be welcome, seeing how applications tend to make some assumptions about these characters anyway, directly or indirectly.

I don't really have a preference. What it really boils down to is the question of "how often are files created on POSIX systems transferred to Windows, and how much of an effort is it for the end user to actually change the name of the file?". As far as I remember, the Windows Explorer at least gives you a nice error message saying that the filename contains invalid characters and doesn't throw any tantrums.

jcdr428:
@xavery very interesting essay from David Wheeler. All is said...
On control characters: "_Control characters in filenames are a terrible idea_"
On metacharacters: "_So, by forbidding a few more characters — at least locally on high-value systems — you eliminate a whole class of programming errors that sometimes become security vulnerabilities".

xavery:
Also, if we *really* want to avoid "bad" characters and file names on Windows, then we should ban all stuff like `CON`, `PRN`, and all that which is mentioned in [Naming Conventions](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions). So the function should be called `IsValidFileName` or even `IsValidPath` and check every component of the pathname against a list of "banned" file/folder names. It would do all this magic mumbo-jumbo on Windows, and pretty much return true on *nix unless there are also some disallowed characters/file names on macOS.

justdan96:
Sorry I've not been as active recently but I can see we have some very healthy discussion! There's certainly a question to be answered on what would be a good idea for filename restrictions. I didn't include the whole path as I just wanted a simple function for filenames that can be expanded on in the future - but as ever nothing is as straightforward as it seems!

I can take a look at including the control characters and Windows naming conventions into the functions - I should have much more time for that next week.

libmediation/types/types.cpp
```diff
@@ -21,6 +22,19 @@
 
 using namespace std;
 
+regex invalidChars()
+{
+#ifndef _WIN32
+    // / and ASCII 0 to 31
+    regex invalid("[/\x00-\x1F]");
```
xavery, line 29: ...you change the `invalid` objects to be `static const`, which means they only get initialised once : there's no need to create a new object every time this function is entered. You will also be able to pass `std::regex_constants::ECMAScript | std::regex_constants::optimize` as the second constructor parameter, potentially instructing the regex engine to construct an object that matches faster at the expense of longer construction time.
I'm not aware of any engines that actually use the `optimize` thing yet, but, well, you know...

libmediation/types/types.cpp
```diff
@@ -350,11 +361,22 @@ string closeDirPath(const string& src, char delimiter)
...
     return src + delimiter;
 }
 
+// extract the filename from a path, check for invalid characters
+bool isValidFileName(const string& src)
+{
+    string filename = extractFileName(src);
+
+    // invalidChars() returns a different regex pattern for Windows or Unix
+    regex ourInvalidChars = invalidChars();
```
xavery, line 375: The type will need to be changed to `auto&&` or `const regex&` if you choose to follow my suggestion about the return value of `invalidChars()`. Or just directly call `invalidChars()` in the `regex_search` call and avoid the temporary variable altogether.

libmediation/types/types.h
```diff
@@ -10,6 +11,8 @@
 #endif
 char* strnstr(const char* s1, const char* s2, size_t len);
 
+std::regex invalidChars();
```
xavery, line 14: There's no need to expose this function : it's entirely internal to the implementation of `types.cpp`.

xavery:
LGTM, have a look at the comments I made though, but there's nothing that serious about them.

justdan96:
Thanks, @xavery I'll have a look at making those changes!

Issue #272, opened 2020-04-05
An issue in the header of wav demuxed from blu-ray m2ts files 


rena-minami:
When I encode the wav demuxed from a bluray pcm file, the flac encoder reports `ERROR: got partial sample`. While using the same track demuxed with eac3to, the encoder can proceed flawlessly.  
Here are the differences highlighted:
<img width="484" alt="Screen Shot 2020-04-05 at 1 24 53 AM" src="https://user-images.githubusercontent.com/63141763/78468406-6e623000-76dd-11ea-9486-1525e0a589d4.png">
<img width="483" alt="Screen Shot 2020-04-05 at 1 25 04 AM" src="https://user-images.githubusercontent.com/63141763/78468408-77eb9800-76dd-11ea-85f8-131dd29577c3.png">
It seems that the "file size" of the header in tsMuxer is filled with `FFFFFF` for both file size (4 bytes after "RIFF") and data size (4 bytes after "data"), (this behavior seems to be hardcoded in line 638 and 655 in `lpcmSteamReader.cpp`). And if I replace the latter `FFFFFF` with the actual size, which is `A0DEC218`, the flac encoder works just fine. 

Meanwhile, the real values for the two `FFFFFF`s, are somehow written at the end of the WAV file, causing it having one more sample than expected. 

If there is any extra information needed, please reply and I will provide accordingly. Thanks.

justdan96:
@jcdr428 can you reproduce this issue?

jcdr428:
@rena-minami the "ffffffff" is normaly replaced at end of reading with file size in method `beforeFileCloseEvent(File& file)` on line 721. It is working as it should on my W64. If you can debug / trace on your system, please see what happens when in this method.

@justdan96 however this has allowed me to discover a new bug in the lpcm to wav, I'll push the fix.

rena-minami:
@jcdr428 Thanks for the reply. After some testing, I can confirm that, the wav extracted on windows is perfectly correct, however, on macOS (10.15.4) and ubuntu (18.04), the issue can be reproduced. I will try to see what happened exactly later today. 
FYI, I found the prebuilt linux binaries does not lunch (directly returns silently), I had to build from source for both tsMuxeR and the GUI.

jcdr428:
@rena-minami so the issue must come from file_linux.cpp . I can't debug linux on my side.

rena-minami:
@jcdr428 I have found what's happening. 

[...]

PR #277, opened 2020-04-10
[bug] Fix wav length

jcdr428:
See issue #272 .


rena-minami:
Is it fine to have two file closes if condition `streamInfo->m_codecReader` is not true? I guess the second one will fail and return false. Or should one of the close be moved inside the if.

jcdr428:
@rena-minami you're right. Moved the close inside the brackets.

PR #297, opened 2020-04-29
Fix location of the main binary in native macOS builds

xavery:


moderndeveloperllc:
@xavery looks to be working - thanks! It didn't occur to me that /Contents/MacOS/ was going to need the CLI, but that makes sense.

xavery:
@moderndeveloperllc Yes, it does. The GUI is completely useless by itself and does not contain any actual code for working with media files.

Issue #317, opened 2020-07-29
disk track information is only partially parsed 


lonecrane:
Hi, I have encountered a blu-ray disk that cannot be parsed correctly by tsmuxer. Its subtitle trackers are not recognized by the latest tsmuxer but BDInfo can show the correct information.
![tsmuxer-1](https://user-images.githubusercontent.com/36885830/88754416-305bb100-d191-11ea-8aa4-3be7a516ff44.png)
![tsmuxer-2](https://user-images.githubusercontent.com/36885830/88754427-35b8fb80-d191-11ea-84f8-8b2c434ae6e6.png)
The following is the zipped file of the disk including the first 32MB of the largest M2TS file, can anyone help to have a look at it?
[http://www.mediafire.com/file/84r03acup7ct4fs/Demux_test.rar/file](url)

xavery:
Unfortunately this doesn't seem resolvable without a full copy of the original BD, and I wasn't able to reproduce this with my own copy, as it's most probably simply a different release.
I'm going to leave this one unassigned if there's anyone else willing to do some work here.

lonecrane:
@xavery Thanks a lot. Actually it is a DIYed BD released by a famous group, and I have encountered another BD with similar problem. Maybe the sample file is so short that it doesn't even contain any subtitle data. I am ready to upload a larger sample or even the full copy if you are willing to give it further study.

jcdr428:
@lonecrane this is not a bug from tsMuxer. The 00000.mpls playlist does not include any subtitle track: when analysing the mpls with an hex editor or BDEdit, you will see that the number of PG files in the STN table is actually 0 !
The subtitle tracks are part of the clpi and m2ts but are "hidden", i.e. not accessible via the Blu-ray menu.

lonecrane:
@jcdr428 Thanks a lot. The new sample including other relevant m2ts files and with a larger size of 128M has been uploaded to
[http://www.mediafire.com/file/h3vnotge4e2xaku/Parasite_2019_B%2526W_Version_Sample-128M.rar/file](url). I am uploading another larger sample in case of any need. When it is ready I will update this reply.

PR #333, opened 2020-09-04
Detect tracks from longest playlist

jcdr428:
tsMuxer currently detects the tracks from the first playlist (i.e. m2ts/clpi) of the mpls.
There can be cases where the subsequent playlists of the mpls contain additional audio/subtitle tracks.
This patch allows tsMuxer to detect the tracks from the playlist having the largest number of tracks.

Solves issue #317.

### DIFF
--- a/tsMuxer/main.cpp
+++ b/tsMuxer/main.cpp
@@ -616,7 +616,7 @@ int main(int argc, char** argv)
                 bool switchToSsif = false;
                 if (mplsParser.m_playItems.size() > 0)
                 {
-                    MPLSPlayItem& item = mplsParser.m_playItems[0];
+                    MPLSPlayItem& item = mplsParser.m_playItems[mplsParser.m_playItem];
                     string itemName = streamDir + item.fileName + mediaExt;
                     if (fileExists(itemName))
                     {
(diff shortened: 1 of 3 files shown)


jcdr428:
@xavery seems I haven't tested enough. The commit provokes crashes in some cases, as reported on doom9:
https://forum.doom9.org/showthread.php?p=1922910#post1922910

I will try to work on this next week end. In the meantime, could you please uncommit and revert the code to c139668 ?
Thanks and sorry for this.

xavery:
@jcdr428 Done. No worries.

ettorps:
Just to let you know: the crashes that guy on the doom9 forum has on Windows (when removing tracks), I have on Linux all the time. Even with builds older than June.
So maybe those crashes are not directly connected to this change.
I wanted to create an issue for this a long time ago, but I always forgot to do it.
When I am back at home today I will open an issue for this.

jcdr428:
Hi @ettorps, thanks for letting us know. It would be nice if you can test previous versions and advise which commit provokes the crashes. With a dichotomy division of time periods it should be quick to find the origin.

lonecrane:
Commit 1c7c911 has worked well on my computer so far. I am running on windows x64 1909.

jcdr428:
@lonecrane I've removed the commit because somebody reported on doom9 that it was causing tsMuxer to crash.
The issue is apparently linked to another older commit, I will push again as soon as the crashing issue is solved.

Issue #188, opened 2020-02-08
UI should be rewritten to use a custom model instead of stuffing everything in a QTableWidget


xavery:
The UI currently uses the `trackLV` member as the central point of information about all the available tracks, and `inputFilesLV` for information on the input files.
This means that a lot of the code in the UI must manually keep the relevant widgets in sync with the information available in the widget. Extending the code in any way easily becomes very cumbersome, as it requires adding custom code for copying data around to other widgets which depend on the data in these `QTableWidget`s. The actual data is contained by `QtvCodecInfo` objects referenced by relevant `QTableWidgetItem`s via a pointer cast to a `qulonglong` and then back.
Since the very early days of Qt 4, the library has supported the [model-view approach](https://doc.qt.io/qt-5/modelview.html) when it comes to supplying data to relevant widgets. The models notify about data being changed via signals, and views update themselves whenever the data changes. This could lead to a significant decrease in the complexity of the UI code, since a lot of work currently done in there would be handled by Qt view classes like `QTableView` and similar. Having a custom model would enable us to use these classes while at the same time providing our own extensions to the API inside our own code.

This would obviously require a lot of work and could potentially introduce breakage. I just wanted to put the idea out there in case anybody else has been wondering about the same thing.

PR #340, opened 2020-09-16
Fix crashes when removing audio/subtitle tracks from the main window

xavery:
Indices in both combo boxes must be updated, since an audio track index might
have changed when a subtitle track is removed and vice versa : there is only one
array which holds information about tracks that these combo boxes reference via
the data contained in their items' UserRoles.


xavery:
The whole "`QComboBox` objects holding onto indices representing private data managed by `QTableWidget`" concept is laughable anyway and causes the whole code to be more complicated than necessary, but we're pretty much stuck with it until somebody takes care of #188 .

PR #376, opened 2020-12-05
Add GUI language

i6henl:
1. Added simplified Chinese language option to GUI.

2. In tsMuxerGUI/lang_codes.h, in order to be more in line with Blu-ray's Chinese DIY habit, modify the "chi" at the top to "zho".

Thank you for your efforts to keep updating tsMuxer!

Best regards！


xavery:
Great work! Thanks for taking the time to modify the code as well.

Issue #374, opened 2020-11-26
tsMuxeR failed to open TS files recorded by nextPVR


bernard0105:
Hi,
I am testing nextPVR to replace WMC in my HTPC. NextPVR records TV shows in .ts files. Those files can be read by VLC, by Kodi etc. Mediainfo can also display the media structure within the files.
I plan to use tsMuxeR as a tool to cut begin and end of the record, and sometime also to join splitted records.
Unfortunately, tsMuxer is failling to open the files.
With those versions tsMuxer_w64-nightly-2020-09-23--02-23-55 and w64-nightly-2020-11-17--01-21-46.zip, it fails without any messages.
With the legacy version tsMuxeR_2.6.12, it fails with a popup error message :
`Error: Bitsream exception Unknown exception. It does not have to be`

Strangely, I have a 5mn records from nextPVR that can be openned by the R2.6.12 but not by the latest versions.
But in that case R2.6.12 failed to open the DVB subtitle tracks, with that Warning message : 
`Some tracks not recognized. This tracks was ignored`
I need to keep the subtitle in the target file.

Maybe it is a single problem related with the DVB subtitle tracks, or maybe it is 2 problems ?
I will put a link to the 5mn records and to a long record.
Thank you.

bernard0105:
Here are the links on pCloud :
- [5mn video](https://u.pcloud.link/publink/show?code=XZd5iRXZz0SYET4zEozgi5cuA3Dei8dKwD0y)
- [long video](https://u.pcloud.link/publink/show?code=XZuJiRXZB3MCabYdaph2xM0W33UrazBmeESV)

jcdr428:
Bonjour @bernard0105, aside from the fact that the detection of DVB subtitles does not work (yet), the error seems to be due to the DVB stream been falsely detected as a hevc stream, resulting in tsMuxer crashing without warning.
Please try again once the fix #380 has been merged, and report.
Merci !

jcdr428:
@bernard0105 please try the latest bintray version and report, thanks.

jcdr428:
@xavery despite @bernard0105 not coming back, I believe the patch fixed the issue and the issue can be closed. @bernard0105 can request to re-open if needed.

PR #380, opened 2020-12-15
Avoid false detection of HEVC

jcdr428:
For detection of each track, tsMuxer parses each stream according to codec specifications.

When parsing a non-hevc stream following hevc specifications, the golomb coding can lead to bitstream exceptions or tsMuxer crashing without warning if the result is above permissible values.
This patch allows to minimize false detections of hevc streams. Eg it should fix issue #374.

### DIFF
--- a/tsMuxer/hevc.cpp
+++ b/tsMuxer/hevc.cpp
@@ -50,6 +50,8 @@ int HevcUnit::deserialize()
         nal_unit_type = m_reader.getBits(6);
         nuh_layer_id = m_reader.getBits(6);
         nuh_temporal_id_plus1 = m_reader.getBits(3);
+        if (nuh_temporal_id_plus1 == 0)
+            return 1;
         return 0;
     }
     catch (BitStreamException& e)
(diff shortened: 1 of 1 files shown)


xavery:
I pushed a small change, since your commit caused the following warnings with Clang and gcc :
```
hevc.cpp: In member function ‘virtual int HevcVpsUnit::deserialize()’:
hevc.cpp:212:53: warning: left shift count >= width of type [-Wshift-count-overflow]
  212 |             if (vps_max_latency_increase_plus1 > (1 << 32) - 2)
      |                                                   ~~^~~~~
hevc.cpp: In member function ‘virtual int HevcSpsUnit::deserialize()’:
hevc.cpp:734:53: warning: left shift count >= width of type [-Wshift-count-overflow]
  734 |             if (sps_max_latency_increase_plus1 > (1 << 32) - 2)
      |                                                   ~~^~~~~
```
Since all platforms that we compile tsMuxer for have an `int` whose width is 32 bits, the expression `1 << 32` is going to overflow, as there is simply no place to store the 32nd bit. Thus, the result of the whole expression becomes implementation-defined, which is something I'd rather avoid.
I think what you mean here was comparing against `0xfffffffe`, since that would be the result of the operation if `1 << 32` was evaluated while automatically being extended to 64 bits. My change assumes that : however, am I correct?

jcdr428:
How are you @xavery ? I did miss this overflow. Comparing against `0xfffffffe` was what I meant.
This makes me realize that (line 18) `HevcUnit::extractUEGolombCode()` really needs to be unsigned, not int.

I will close this pull and work on it further.

xavery:
I'm doing well, thanks. :-) No need to close the PR outright : there are lots of places that erroneously use signed integers for bit operations anyway. If you want to work some more on this, that's fine, but these changes are okay to me as far as I'm concerned. Let me know if you change your mind and I'll merge this.

jcdr428:
@xavery ok, let's merge this.

PR #382, opened 2020-12-19
Avoid bitstream exceptions in AVC deserializer

jcdr428:
When parsing a stream other than AVC with the AVC deserializer, allow early return and avoid false AVC detection and bitstream exceptions.
Plus correction of `ExtractUEGolombCode ` from int to unsigned types.

### DIFF
--- a/tsMuxer/h264StreamReader.cpp
+++ b/tsMuxer/h264StreamReader.cpp
@@ -1367,9 +1367,9 @@ int H264StreamReader::processSPS(uint8_t* buff)
     if (replaceToOwnSPS())
     {
         // do not updat SPS in stream because of we are going to insert own sps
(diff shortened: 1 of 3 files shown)


jcdr428:
@xavery would it be better to use `size_t` rather than `unsigned` for ExtractUEGolombCode objects ?

xavery:
> @xavery would it be better to use `size_t` rather than `unsigned` for ExtractUEGolombCode objects ?

I don't really have a preference. There's too many `int`s around anyway, and bitwise values should generally be expressed as `unsigned`. In a perfect world, this would be a `std::bitset<N>` with `N` being a template parameter of the function that produces it, expressing how many bits are expected to be read from the given reader object.

PR #384, opened 2020-12-21
CMake tweaks to make the project more linux packager friendly

hobbes1069:
- Add install target for tsMuxerGUI
- Add install for contributed desktop file
- Install icon for desktop file
- Don't check for C when all sources are C++.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,13 +23,15 @@ else()
 endif()
 add_definitions("-DTSMUXER_VERSION=\"${TSMUXER_VERSION}\"")
 
-if(UNIX)
+if(NOT MSVC)
   include(GNUInstallDirs)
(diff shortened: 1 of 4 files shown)


qyot27:
`if(UNIX)` should be `if(NOT MSVC)`, so as not to exclude MinGW-w64 (either inside of MSys2 or when cross-compiling).

hobbes1069:
Since install targets were not included for tsMuxerGUI I decided to "play it safe" but of course could be updated to include more targets.

xavery:
Great work! Thank you very much.
I was worried that changing the install location might result in the nightly builds breaking, but CMake documentation says that `CMAKE_INSTALL_BINDIR` is `bin` by default, so it looks like this is going to be just fine.

PR #385, opened 2020-12-22
Fix signed/unsigned mismatch warning messages

jcdr428:
Fix all signed/unsigned mismatch in `<` `>` comparisons.

### DIFF
--- a/tsMuxer/aacStreamReader.cpp
+++ b/tsMuxer/aacStreamReader.cpp
@@ -9,7 +9,7 @@
 #include "vodCoreException.h"
 #include "vod_common.h"
 
-unsigned AACStreamReader::getHeaderLen() { return AAC_HEADER_LEN; }
+int AACStreamReader::getHeaderLen() { return AAC_HEADER_LEN; }
 
 const std::string AACStreamReader::getStreamInfo()
 {
--- a/tsMuxer/ac3Codec.cpp
+++ b/tsMuxer/ac3Codec.cpp
@@ -119,8 +119,8 @@ int AC3Codec::parseHeader(uint8_t* buf, uint8_t* end)
     m_bsid = id;
     if (m_bsid > 10)  // bsid = 16 => EAC3
     {
-        unsigned int numblkscod, strmtyp, substreamid, number_of_blocks_per_syncframe;
-        uint8_t acmod, lfeon, bsmod, fscod, dsurmod, pgmscle, extpgmscle, mixdef, paninfoe;
+        int numblkscod, strmtyp, substreamid, number_of_blocks_per_syncframe;
+        int acmod, lfeon, bsmod, fscod, dsurmod, pgmscle, extpgmscle, mixdef, paninfoe;
         bsmod = fscod = dsurmod = pgmscle = extpgmscle = mixdef = paninfoe = 0;
 
         strmtyp = gbc.getBits(2);
--- a/tsMuxer/blurayHelper.cpp
+++ b/tsMuxer/blurayHelper.cpp
@@ -449,11 +449,11 @@ bool BlurayHelper::createCLPIFile(TSMuxer* muxer, int clpiNum, bool doLog)
         CLPIStreamInfo streamInfo(itr->second);
         clpiParser.m_streamInfo.insert(make_pair(streamInfo.streamPID, streamInfo));
     }
-    vector<int64_t> packetCount = muxer->getMuxedPacketCnt();
+    vector<uint64_t> packetCount = muxer->getMuxedPacketCnt();
     vector<int64_t> firstPts = muxer->getFirstPts();
     vector<int64_t> lastPts = muxer->getLastPts();
 
-    for (unsigned i = 0; i < muxer->splitFileCnt(); i++)
+    for (size_t i = 0; i < muxer->splitFileCnt(); i++)
     {
         if (muxer->isInterleaveMode())
             clpiParser.interleaveInfo = muxer->getInterleaveInfo(i);
@@ -556,7 +556,7 @@ bool BlurayHelper::createMPLSFile(TSMuxer* mainMuxer, TSMuxer* subMuxer, int aut
     }
     else
     {
-        for (int i = 0; i < customChapters.size(); i++)
+        for (size_t i = 0; i < customChapters.size(); i++)
         {
             int64_t mark = customChapters[i] * 45000.0;
             if (mark >= 0 && mark <= (mplsParser.OUT_time - mplsParser.IN_time))
--- a/tsMuxer/blurayHelper.cpp
+++ b/tsMuxer/blurayHelper.cpp
@@ -556,7 +556,7 @@ bool BlurayHelper::createMPLSFile(TSMuxer* mainMuxer, TSMuxer* subMuxer, int aut
     }
     else
     {
-        for (auto &i : customChapters)
+        for (auto& i : customChapters)
         {
             int64_t mark = i * 45000.0;
             if (mark >= 0 && mark <= (mplsParser.OUT_time - mplsParser.IN_time))
--- a/tsMuxer/blurayHelper.cpp
+++ b/tsMuxer/blurayHelper.cpp
@@ -556,9 +556,9 @@ bool BlurayHelper::createMPLSFile(TSMuxer* mainMuxer, TSMuxer* subMuxer, int aut
     }
     else
     {
-        for (size_t i = 0; i < customChapters.size(); i++)
+        for (auto &i : customChapters)
         {
-            int64_t mark = customChapters[i] * 45000.0;
+            int64_t mark = i * 45000.0;
             if (mark >= 0 && mark <= (mplsParser.OUT_time - mplsParser.IN_time))
                 mplsParser.m_marks.push_back(PlayListMark(-1, mark + mplsParser.IN_time));
         }
--- a/tsMuxer/bufferedReaderManager.cpp
+++ b/tsMuxer/bufferedReaderManager.cpp
@@ -27,7 +27,14 @@ void BufferedReaderManager::init(uint32_t blockSize, uint32_t allocSize, uint32_
     m_prereadThreshold = prereadThreshold > 0 ? prereadThreshold : m_blockSize / 2;
 }
 
-BufferedReaderManager::~BufferedReaderManager() { std::vector<BufferedReader*>().swap(m_fileReaders); }
+BufferedReaderManager::~BufferedReaderManager()
+{
+    for (size_t i = 0; i < m_fileReaders.size(); i++)
+    {
+        delete m_fileReaders[i];  // need to define destruction order first. This object MUST be deleted after
+                                  // MCVodStreamer
+    }
+}
 
 AbstractReader* BufferedReaderManager::getReader(const char* streamName)
 {
--- a/tsMuxer/bufferedReaderManager.cpp
+++ b/tsMuxer/bufferedReaderManager.cpp
@@ -29,11 +29,7 @@ void BufferedReaderManager::init(uint32_t blockSize, uint32_t allocSize, uint32_
 
 BufferedReaderManager::~BufferedReaderManager()
 {
-    for (size_t i = 0; i < m_fileReaders.size(); i++)
-    {
-        delete m_fileReaders[i];  // need to define destruction order first. This object MUST be deleted after
-                                  // MCVodStreamer
-    }
+    m_fileReaders.clear();
 }
 
 AbstractReader* BufferedReaderManager::getReader(const char* streamName)
--- a/tsMuxer/bufferedReaderManager.cpp
+++ b/tsMuxer/bufferedReaderManager.cpp
@@ -27,10 +27,7 @@ void BufferedReaderManager::init(uint32_t blockSize, uint32_t allocSize, uint32_
     m_prereadThreshold = prereadThreshold > 0 ? prereadThreshold : m_blockSize / 2;
 }
 
-BufferedReaderManager::~BufferedReaderManager()
-{
-    m_fileReaders.clear();
-}
+BufferedReaderManager::~BufferedReaderManager() { m_fileReaders.clear(); }
 
 AbstractReader* BufferedReaderManager::getReader(const char* streamName)
 {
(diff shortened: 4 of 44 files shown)


xavery:
Ugh, this is big ... and might lead to issues where some code unknowingly relying on signed overflow behaviour (for example), or the integers' signedness in itself.
I'm going to leave some inline comments in the code where I think it should be done in a different way.
Generally, I would be against merging this and just let these warnings be, as the potential dangers are IMHO far worse.

tsMuxer/aacStreamReader.cpp
```diff
@@ -9,7 +9,7 @@
 #include "vodCoreException.h"
 #include "vod_common.h"
 
-unsigned AACStreamReader::getHeaderLen() { return AAC_HEADER_LEN; }
+int AACStreamReader::getHeaderLen() { return AAC_HEADER_LEN; }
```
xavery, line 12: If this is an actual length, it shouldn't be negative : if anything, it's `AAC_HEADER_LEN`'s type that should be changed to unsigned.

tsMuxer/ac3Codec.cpp
```diff
@@ -119,8 +119,8 @@ int AC3Codec::parseHeader(uint8_t* buf, uint8_t* end)
     m_bsid = id;
     if (m_bsid > 10)  // bsid = 16 => EAC3
     {
-        unsigned int numblkscod, strmtyp, substreamid, number_of_blocks_per_syncframe;
-        uint8_t acmod, lfeon, bsmod, fscod, dsurmod, pgmscle, extpgmscle, mixdef, paninfoe;
+        int numblkscod, strmtyp, substreamid, number_of_blocks_per_syncframe;
+        int acmod, lfeon, bsmod, fscod, dsurmod, pgmscle, extpgmscle, mixdef, paninfoe;
```
xavery, line 123: These variables originate from the bitstream itself (as far as I see), thus they're most probably bitfields and should be treated as unsigned.

tsMuxer/h264StreamReader.cpp
```diff
@@ -243,7 +243,7 @@ int H264StreamReader::writeSEIMessage(uint8_t* dstBuffer, uint8_t* dstEnd, SEIUn
         static const uint8_t DEFAULT_MVC_SEI_HEADER[] = {192, 16};
         if (!m_lastSeiMvcHeader.empty())
         {
-            for (int i = 0; i < m_lastSeiMvcHeader.size(); ++i) writer.putBits(8, m_lastSeiMvcHeader[i]);
+            for (size_t i = 0; i < m_lastSeiMvcHeader.size(); ++i) writer.putBits(8, m_lastSeiMvcHeader[i]);
```
xavery, line 246: Can be replaced with a ranged-for loop.

tsMuxer/iso_writer.cpp
```diff
@@ -333,8 +333,8 @@ bool FileEntryInfo::isFile() const { return m_fileType == FileType_File || m_fil
 FileEntryInfo::~FileEntryInfo()
 {
     delete[] m_sectorBuffer;
-    for (int i = 0; i < m_files.size(); ++i) delete m_files[i];
-    for (int i = 0; i < m_subDirs.size(); ++i) delete m_subDirs[i];
+    for (size_t i = 0; i < m_files.size(); ++i) delete m_files[i];
+    for (size_t i = 0; i < m_subDirs.size(); ++i) delete m_subDirs[i];
```
xavery, line 337: ranged-for

tsMuxer/tsMuxer.cpp
```diff
@@ -507,7 +507,7 @@ bool TSMuxer::close()
 int TSMuxer::calcM2tsFrameCnt()
 {
     uint32_t byteCnt = 0;
-    for (int i = 0; i < m_m2tsDelayBlocks.size(); i++) byteCnt += m_m2tsDelayBlocks[i].second;
+    for (size_t i = 0; i < m_m2tsDelayBlocks.size(); i++) byteCnt += m_m2tsDelayBlocks[i].second;
```
xavery, line 510: ranged-for

tsMuxer/tsMuxer.cpp
```diff
@@ -645,7 +645,7 @@ void TSMuxer::buildPesHeader(int pesStreamID, AVPacket& avPacket, int pid)
     int bufLen = pesPacket->getHeaderLength() + additionDataSize;
     m_pesData.resize(bufLen);
     memcpy(m_pesData.data(), tmpBuffer, bufLen);
-    for (int i = 0; i < tmpPriorityData.size(); ++i)
+    for (size_t i = 0; i < tmpPriorityData.size(); ++i)
```
xavery, line 648: ranged-for

xavery:
There's generally a lot of places where ranged-for can be used, thus nullifying the need for an index variable altogether. If you really want to go through with this, I'd recommend changing these loops to use the "new" form if the index variable isn't used in the loop body.

tsMuxer/iso_writer.cpp
jcdr428: Wouldn't the following do the same thing:
```
m_files.clear();
m_subDirs.clear();
```

tsMuxer/h264StreamReader.cpp
jcdr428: Done.

tsMuxer/tsMuxer.cpp
jcdr428: Done.

tsMuxer/tsMuxer.cpp
jcdr428: Done.

tsMuxer/aacStreamReader.cpp
jcdr428: If getHeaderLen() is changed to unsigned type, then we would have to cast e.g. `if (m_bufEnd - m_curPos < getHeaderLen())` and other variables.

There is no real reason to use an unsigned type for audio header length `getHeaderLen()`, as the value is guaranteed to be below 14: returned values LPCM = 4, AAC = 7, AC3 = 7, DVB = 10, DTS = 14

tsMuxer/ac3Codec.cpp
jcdr428: No real reason to use `unsigned`:
`numblkscod` is 2 bits, `strmtyp` is 2 bits, `substreamid` is 3 bits, `number_of_blocks_per_syncframe` is <6, `acmod` is 3 bits, `lfeon` is 1 bit, `bsmod` is 3 bits, `fscod` is 2 bits, `dsurmod` is 2 bits, `pgmscle` is 1 bit, `extpgmscle` is 1 bit, `mixdef` is 2 bits, `paninfoe` is 1 bit.

But if you think `unsigned `is better than `int` in this case, I'll change.

tsMuxer/tsMuxer.cpp
```diff
@@ -1400,9 +1399,9 @@ void TSMuxer::parseMuxOpt(const std::string& opts)
         {
             uint64_t coeff = 1;
             string postfix;
-            for (int i = paramPair[1].size() - 1; i >= 0; i--)
```
jcdr428, line 1403: Bug: double `for i =` loop

tsMuxer/textSubtitlesRender.cpp
```diff
@@ -386,12 +386,12 @@ bool TextSubtitlesRender::rasterText(const std::string& text)
     vector<string> lines = splitStr(text.c_str(), '\n');
     int curY = 0;
     m_initFont = m_font;
-    for (int i = 0; i < lines.size(); ++i)
+    for (auto& i : lines)
     {
-        vector<pair<Font, string>> txtParts = processTxtLine(lines[i], fontStack);
-        for (int i = 0; i < txtParts.size(); ++i)
```
jcdr428, line 392: Bug: double `for i =` loop

tsMuxer/dvbSubStreamReader.cpp
```diff
@@ -38,8 +38,8 @@ int DVBSubStreamReader::intDecodeFrame(uint8_t* buff, uint8_t* end)
 {
     return 0;  // not working yet
 
-    unsigned pos, cmd, x1, y1, x2, y2, offset1, offset2, next_cmd_pos;
-    unsigned cmd_pos, is_8bit = 0;
+    int pos, cmd, x1, y1, x2, y2, offset1, offset2, next_cmd_pos;
+    int cmd_pos, is_8bit = 0;
```
jcdr428, line 42: Note that this code is never executed (`return 0;  // not working yet`).
This code is a copy of [ffmpeg DVD subtitle decoder](https://github.com/FFmpeg/FFmpeg/blob/2f20e35892c866d4e141e73a3e2df16457f9a21a/libavcodec/dvdsubdec.c#L220), nothing to do with DV**B** subtitles. Maybe it is better to remove to avoid confusion and comply with GNU license.

tsMuxer/textSubtitles.h
```diff
@@ -60,7 +60,7 @@ class TextToPGSConverter  //: public TextSubtitlesRenderWin32
     void setImageBuffer(uint8_t* value) { m_imageBuffer = value; }
 
    public:
-    uint32_t m_rleLen;
+    int m_rleLen;
```
jcdr428, line 63: `rleLen` maximum is image width + 16

tsMuxer/blurayHelper.cpp
```diff
@@ -449,11 +449,11 @@ bool BlurayHelper::createCLPIFile(TSMuxer* muxer, int clpiNum, bool doLog)
         CLPIStreamInfo streamInfo(itr->second);
         clpiParser.m_streamInfo.insert(make_pair(streamInfo.streamPID, streamInfo));
     }
-    vector<int64_t> packetCount = muxer->getMuxedPacketCnt();
+    vector<uint64_t> packetCount = muxer->getMuxedPacketCnt();
     vector<int64_t> firstPts = muxer->getFirstPts();
     vector<int64_t> lastPts = muxer->getLastPts();
```
jcdr428, line 454: PTS is 32-bit in the PES packet.
`nanoClockToPTS()` returns a `int64_t` type.

tsMuxer/hevc.h
```diff
@@ -212,7 +212,7 @@ struct HevcSliceHeader : public HevcUnit
     bool first_slice;
     unsigned pps_id;
     unsigned slice_type;
-    unsigned pic_order_cnt_lsb;
+    int pic_order_cnt_lsb;
```
jcdr428, line 215: `pic_order_cnt_lsb` is 16-bit max

tsMuxer/bufferedReaderManager.cpp
```diff
@@ -27,14 +27,7 @@ void BufferedReaderManager::init(uint32_t blockSize, uint32_t allocSize, uint32_
...
 
-BufferedReaderManager::~BufferedReaderManager()
-{
-    for (size_t i = 0; i < m_fileReaders.size(); i++)
-    {
-        delete m_fileReaders[i];  // need to define destruction order first. This object MUST be deleted after
-                                  // MCVodStreamer
-    }
-}
+BufferedReaderManager::~BufferedReaderManager() { std::vector<BufferedReader*>().swap(m_fileReaders); }
```
xavery, line 30: This will cause memory leaks, since the `BufferedReader` objects inside `m_fileReaders` will never be destructed, as the vector consists of raw pointers, not `unique_ptr`s or similar objects which also manage the object's lifetime.
Since it's not straightforward to just change `BufferedReader*` to `unique_ptr<BufferedReader>` (which would certainly be welcome but it's definitely out of the scope of this PR), I'm afraid this loop must stay here.

tsMuxer/metaDemuxer.cpp
```diff
@@ -287,13 +287,10 @@ int METADemuxer::addPGSubStream(const string& codec, const string& _codecStreamN
...
-    for (int i = 0; i < mplsInfoList.size(); ++i)
+    for (auto& i : mplsInfoList)
     {
-        const MPLSParser& mplsInfo = mplsInfoList[i];
-        for (int j = 0; j < mplsInfo.m_playItems.size(); ++j)
-        {
-            result.push_back(mplsInfo.m_playItems[j]);
-        }
+        const MPLSParser& mplsInfo = i;
+        for (auto& j : mplsInfo.m_playItems) result.push_back(j);
```
xavery, line 293: Could be just `for (auto& j : i.m_playItems)`, it seems, no need for the additional variable.

tsMuxer/srtStreamReader.cpp
```diff
@@ -192,7 +192,7 @@ uint8_t* SRTStreamReader::renderNextMessage(uint32_t& renderedLen)
         m_state = PARSE_TIME;
         bool isNUmber = true;
         {
-            for (auto c : m_sourceText.front())
+            for (auto& c : m_sourceText.front())
```
xavery, line 195: Just out of curiosity, what's the reason behind this? There doesn't seem to be any benefit to accessing this container via reference vs. via value.

jcdr428: Just to be consistent with all other ranged-based for loops: there is no need to copy the value to another variable here.

tsMuxer/metaDemuxer.cpp
jcdr428: Done.

tsMuxer/bufferedReaderManager.cpp
jcdr428: Ok, reverted.

tsMuxer/iso_writer.cpp
```diff
@@ -333,8 +333,8 @@ bool FileEntryInfo::isFile() const { return m_fileType == FileType_File || m_fil
 FileEntryInfo::~FileEntryInfo()
 {
     delete[] m_sectorBuffer;
-    for (int i = 0; i < m_files.size(); ++i) delete m_files[i];
-    for (int i = 0; i < m_subDirs.size(); ++i) delete m_subDirs[i];
+    std::vector<FileEntryInfo*>().swap(m_files);
+    std::vector<FileEntryInfo*>().swap(m_subDirs);
```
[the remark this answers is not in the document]
jcdr428: Ok, reverted.

tsMuxer/srtStreamReader.cpp
xavery: I was going to say "but it's potentially more costlier to use a reference here", but a quick peek at [godbolt](https://gcc.godbolt.org/z/sG84qT) proved me wrong : the code is literally the same in both cases.

PR #394, opened 2021-01-23
tsMuxerGUI CMakeLists fixes

nu774:
* add /utf-8 on MSVC compiler
* add WIN32 or MACOSX_BUNDLE to add_executable()


xavery:
Thanks!

PR #399, opened 2021-02-07
Nightlies on GitHub

### LINKED ISSUE #397: Bintray is shutting down
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Bintray will not be accepting any new submissions (I'm assuming this also means new package versions) after the 28th of February.
We should probably move to use GitHub Releases for hosting the nightly builds : appropriate actions (for example [create-release](https://github.com/actions/create-release)) in the workflows can be configured to create a new release and upload the binaries. I have investigated this a bit when refactoring the nightly scripts, but finally decided to stick to Bintray since there were no real reasons not to.

Unless somebody has any objections or other suggestions, I'll work towards this in the following weeks, and make an official announcement on Doom9 when the move is done.

xavery:
Rework the nightly actions workflow in order to store nightlies in GitHub

Fixes #397 .


xavery:
I tested this in my [playground repo](https://github.com/xavery/actions-playground) by using the same actions and replicating the workflow as much as possible. It worked, so I'm merging this hoping that it does work here as well.
However, I never seem to get these things right the first time anyway, so trivial fixes tomorrow are to be expected.

Issue #409, opened 2021-04-25
Crash of tsmuxer of hevcStreamReader.cpp in function HEVCStreamReader::updateStreamFps


NigelX:
Hi
I found an crash error.

System info：
Ubuntu 20.04 : clang 10.0.0 , gcc 9.3.0

tsMuxeR version git-a50c2c5

------------------------------------------
Verification steps：
1.Get the source code of tsMuxer
2.Compile the tsMuxer
```bash
$ cd tsMuxer
$ mkdir build && cd build
$ cmake ../ -DCMAKE_C_COMPILER=clang  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_FLAGS="-fsanitize=address" -DCMAKE_CXX_FLAGS="-fsanitize=address"
$ make -j 32
```
3.run poc
```bash
$ cd tsMuxer
$ ./tsmuxer poc
```
file: [poc.zip](https://github.com/justdan96/tsMuxer/files/6371318/1.tsmuxer_HEVCStreamReader_poc.wav.zip)

asan info:
```bash
tsMuxeR version git-a50c2c5. github.com/justdan96/tsMuxer
HEVC muxing fps is not set. Get fps from stream. Value: 2.38528
HEVC manual defined fps doesn't equal to stream fps. Change HEVC fps from 0.50014 to 2.38528
=================================================================
==1586847==ERROR: AddressSanitizer: negative-size-param: (size=-139729651620667)
    #0 0x5573cc in __asan_memmove (/home/hh/Downloads/tsMuxer/build/tsMuxer/tsmuxer+0x5573cc)
    #1 0x650650 in HEVCStreamReader::updateStreamFps(void*, unsigned char*, unsigned char*, int) /home/hh/Downloads/tsMuxer/tsMuxer/hevcStreamReader.cpp:375:9
    #2 0x7d07ab in MPEGStreamReader::updateFPS(void*, unsigned char*, unsigned char*, int) /home/hh/Downloads/tsMuxer/tsMuxer/mpegStreamReader.cpp:310:9
    #3 0x64a214 in HEVCStreamReader::checkStream(unsigned char*, int) /home/hh/Downloads/tsMuxer/tsMuxer/hevcStreamReader.cpp:68:17
[...]

PR #412, opened 2021-04-25
[bug] Avoid memory overflow when m_bufEnd not set

jcdr428:
Solves issue #409 .


xavery:
Verified with clang + asan.

Issue #411, opened 2021-04-25
heap-buffer-overflow of tsmuxer of hevc.cpp in function HevcSpsUnit::short_term_ref_pic_set(int) 


NigelX:
Hi
I found an crash error.

System info：
Ubuntu 20.04 : clang 10.0.0 , gcc 9.3.0

tsMuxeR version git-a50c2c5

commit：a50c2c5a169546beaa6fff93f1be785b3b8aed68

------------------------------------------
Verification steps：
1.Get the source code of tsMuxer
2.Compile the tsMuxer
```bash
$ cd tsMuxer
$ mkdir build && cd build
$ cmake ../ -DCMAKE_C_COMPILER=clang  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_FLAGS="-fsanitize=address" -DCMAKE_CXX_FLAGS="-fsanitize=address"
$ make -j 32
```
3.run poc
```bash
$ cd tsMuxer
$ ./tsmuxer poc
```
file:
[poc.zip](https://github.com/justdan96/tsMuxer/files/6371327/3.tsmuxerHevcSpsUnit_short_term_ref_pic_set_poc.wav.zip)

asan info:
```bash
tsMuxeR version git-a50c2c5. github.com/justdan96/tsMuxer
=================================================================
==2094991==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x617000002d10 at pc 0x00000063bb0b bp 0x7ffc52150a70 sp 0x7ffc52150a68
WRITE of size 4 at 0x617000002d10 thread T0
    #0 0x63bb0a in HevcSpsUnit::short_term_ref_pic_set(int) /home/hh/Downloads/tsMuxer/tsMuxer/hevc.cpp:578:35
    #1 0x63df14 in HevcSpsUnit::deserialize() /home/hh/Downloads/tsMuxer/tsMuxer/hevc.cpp:869:17
    #2 0x64a32c in HEVCStreamReader::checkStream(unsigned char*, int) /home/hh/Downloads/tsMuxer/tsMuxer/hevcStreamReader.cpp:74:24
    #3 0x75372b in METADemuxer::detectTrackReader(unsigned char*, int, AbstractStreamReader::ContainerType, int, int) /home/hh/Downloads/tsMuxer/tsMuxer/metaDemuxer.cpp:770:21
[...]

PR #413, opened 2021-04-25
[bug] Limit number of pos. or neg. pics

jcdr428:
Early abort when num_negative_pics or nb_positive_pics > 64 (max number of ref. frames).

Solves issue #411 .


xavery:
Verified with Clang + asan.

Issue #417, opened 2021-05-19
heap-buffer-overflow of bitStream.h in function getCurVal


NigelX:
Hi
I found an crash error.

System info：
Ubuntu 20.04 : clang 10.0.0 , gcc 9.3.0

tsMuxeR version git-5a9adef

------------------------------------------
Verification steps：
1.Get the source code of tsMuxer
2.Compile the tsMuxer
```bash
$ cd tsMuxer
$ mkdir build && cd build
$ cmake ../ -DCMAKE_C_COMPILER=clang  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_FLAGS="-fsanitize=address" -DCMAKE_CXX_FLAGS="-fsanitize=address"
$ make -j 32
```
3.run poc
```bash
$ cd tsMuxer
$ ./tsmuxer poc
```
file:
[poc.zip](https://github.com/justdan96/tsMuxer/files/6507109/poc1.zip)

info:
```bash
Program received signal SIGFPE, Arithmetic exception.
[----------------------------------registers-----------------------------------]
RAX: 0x4a817c80000 
RBX: 0x7ffff3814013 --> 0x676964780400770b 
RCX: 0x0 
RDX: 0x0 
RSI: 0x3 
RDI: 0x7fffffff79a0 --> 0x100000048 
RBP: 0x7 
RSP: 0x7fffffff79a0 --> 0x100000048 
RIP: 0x28f747 (<AC3Codec::decodeDtsHdFrame(unsigned char*, unsigned char*)+1767>:	idiv   rcx)
R8 : 0x48 ('H')
R9 : 0x0 
R10: 0x666eb0 --> 0x0 
R11: 0x5354524d44415440 ('@TADMRTS')
R12: 0x7fffffff82a8 --> 0x0 
R13: 0x7ffff381412c --> 0x0 
R14: 0x7fffffff82a8 --> 0x0 
R15: 0xa ('\n')
EFLAGS: 0x10246 (carry PARITY adjust ZERO sign trap INTERRUPT direction overflow)
[-------------------------------------code-------------------------------------]
   0x28f73b <AC3Codec::decodeDtsHdFrame(unsigned char*, unsigned char*)+1755>:	shr    rdx,0x20
   0x28f73f <AC3Codec::decodeDtsHdFrame(unsigned char*, unsigned char*)+1759>:	je     0x28f821 <AC3Codec::decodeDtsHdFrame(unsigned char*, unsigned char*)+1985>
[...]

PR #421, opened 2021-05-19
[bug] Division by 0

jcdr428:
Early return when `mh.group1_samplerate` (Sample rate of first substream) is 0, to avoid division by 0 error.

Solves issues #417 and #428.


justdan96:
Are there any knock-on effects to the early return?

jcdr428:
@justdan96 there shouldn't be: if the read sample rate is zero, then it means that the stream is not a MPL and the MPL parsing should be aborted.

PR #449, opened 2021-09-01
Reduce OSXcross Build Dependencies

justdan96:
This, along with https://github.com/justdan96/tsmuxer_build/pull/4 will reduce build dependencies for the OSXcross build, so that those builds no longer need to install additional dependencies from Homebrew.


justdan96:
merging despite code style check failure

PR #470, opened 2021-09-29
Qt6 Readiness

justdan96:
This PR is to work on Qt6 readiness. Currently there is only Qt 6.1.3 available and it does not include QtMultimedia but 6.2.0 should be released soon and that will include it.

To be ready I have updated some GUI code to work in Qt6 - it should be backwards-compatible - and added the possibility to not use QtMultimedia, just in case it is useful.

### DIFF
--- a/.github/workflows/qt6_test.sh
+++ b/.github/workflows/qt6_test.sh
@@ -1,6 +1,15 @@
 cd tsMuxerGUI
+
 rm -fR build
 mkdir build
 cd build
 qmake ..
 make -j$(nproc --all)
+
+cd ..
+
+rm -fR build
+mkdir build
+cd build
+cmake -DTSMUXER_GUI=TRUE -DQT_VERSION=6 -GNinja ..
+ninja tsMuxerGUI
--- a/.github/workflows/qt6_test.sh
+++ b/.github/workflows/qt6_test.sh
@@ -11,5 +11,5 @@ cd ..
 rm -fR build
 mkdir build
 cd build
-cmake -DTSMUXER_GUI=TRUE -DQT_VERSION=6 -GNinja ..
-ninja tsMuxerGUI
+cmake -DTSMUXER_GUI=TRUE -DQT_VERSION=6 ..
+make -j$(nproc --all) tsMuxerGUI
--- a/.github/workflows/qt6_test.sh
+++ b/.github/workflows/qt6_test.sh
@@ -1,15 +1,13 @@
 cd tsMuxerGUI
 
-rm -fR build
-mkdir build
-cd build
+mkdir qmake_build
+cd qmake_build
 qmake ..
 make -j$(nproc --all)
 
-cd ..
+cd ../..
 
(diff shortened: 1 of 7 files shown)


justdan96:
Just a note - the last version of Qt supported by XP was 5.6.3. The "Qt 5.6.3 for XP without SSE" build that Xavery created is effectively the latest one that will work. In order to proceed we would need to make some changes just to support that build, specifically `tabStopDistance` replaces `tabStopWidth` since 5.10, so that would have to be amended only for the XP build.

xavery:
While I'm all for switching to newer versions of Qt, keep in mind that Qt 5 is the last version to officially support Windows 7. Qt 6 [lists Windows 10](https://doc.qt.io/qt-6/supported-platforms.html#windows) as the earliest supported Windows version.

Seeing how we're still providing Windows XP builds despite the OS being 20 years old this year, I'm going to bet that a significant part of tsMuxer's user base still uses Windows 7.

Also, I managed to create a codebase that's compilable with both Qt 5.6.3 and Qt 6.2 in [one of my other projects](https://github.com/xavery/czateria/commit/2c1e302c087e6e8a12fea1b29ac9f96403f8e521), though with a couple `#if QT_VERSION` checks here and there. Perhaps the same will be possible here.

justdan96:
@xavery thanks for the link, I'll look at making sure it is backwards compatible with Qt5 as well.

xavery:
> @xavery thanks for the link, I'll look at making sure it is backwards compatible with Qt5 as well.

Actually, I have some changes ready : should be pushing them soon.

xavery:
Alright, here's a summary : 

* `QProcess::error` and `BackgroundColorRole` were already deprecated in Qt 5.6, so there's no harm in switching to `errorOccurred` and `BackgroundRole` respectively.
* `QVariant::fromValue` is also available in Qt 5.6, so there's no point to using `QVariant::setValue` which is also longer since the variant object must be created beforehand.
* `QSet::toList` was removed in Qt 6. I tried using the `QList`'s iterator-based constructor instead, but it was added in Qt 5.14. Switching to standard library containers seems like a good solution here, as there's no real reason to use `QSet` or `QList` in this case.
* `setTabStopDistance` / `setTabStopWidth` was the worst of it all, as UI files have no facilities to incorporate Qt version differences, at least none that I could find. The solution is to `define` the appropriate override before including the generated header if using pre-5.10 Qt.
* The `setIniCodec` call must be preserved in pre-6 Qt versions as they don't have a default character encoding for INI files, which means that some versions/locales might use non-UTF-8 encoding and break compatibility.

I also incorporated your changes related to skipping QtMultimedia parts if the module is not available.

Also, there's a new workflow which attempts a Qt 6.2.0 build. This way we can immediately see if the code is compatible with both 5.6.3 and 6.2.0.

justdan96:
Thanks for the changes, the modifications are looking a lot better now! All the checks are passing so I think we are probably okay to merge this one.

PR #485, opened 2021-11-01
Use the model-view approach for handling font settings

xavery:
This change moves all code related to handling font settings from TsMuxerWindow to its own model. The widget handling the data is now just an ordinary QTableView.

This does away with storing everything as strings inside a QTableWidget, which means that functions using font data to generate metafile or other kinds of output use QFont objects directly.


justdan96:
Wow this is great - thanks a lot @xavery !

PR #488, opened 2021-11-08
Use -D_FILE_OFFSET_BITS=64 instead of offset-width-specific functions

xavery:


justdan96:
Thanks a lot for this!

PR #491, opened 2021-11-11
Split directory.cpp into unix/win32 parts

xavery:
`directory.cpp` was essentially two implementations in one, separated with `ifdef`s. It's cleaner to just have two separate files, seeing how `CMakeLists.txt` for `libmediation` needs to check the platform it's running on and add the appropriate `file` implementation anyway.

This commit splits this file into two and extracts common code parts to `directory_priv.h`.


justdan96:
Thanks a lot for this, it really does make a lot more sense to have it split!

PR #492, opened 2021-11-12
Remove #ifdef apple in order to use std::filesystem

xavery:
It seems like `std::__fs` is a leftover from the days when support for `filesystem` was still experimental in AppleClang. `filesystem` was officially declared as supported in Xcode 11 when targeting macOS 10.15, which is our current `MACOSX_DEPLOYMENT_TARGET`, so there is no need to keep this `ifdef` around.


jcdr428 (approved):
@xavery indeed, no need for `std::__fs` anymore.

PR #496, opened 2021-11-21
Use static freetype from Homebrew in MacOS native builds

xavery:


justdan96:
I'll try to find some time this weekend to try this out on an M1 Mac.

build_macos_native.sh
```diff
@@ -5,15 +5,14 @@ set -e
...
 brew install freetype
 
-builddir=$PWD
-
 mkdir build
 
 pushd build
-cmake -DCMAKE_BUILD_TYPE=Release -DTSMUXER_GUI=TRUE ..
+cmake -DCMAKE_BUILD_TYPE=Release -DTSMUXER_STATIC_BUILD=TRUE \
+  '-DFREETYPE_LDFLAGS=bz2;/usr/local/lib/libpng.a' -DTSMUXER_GUI=TRUE \
```
justdan96, line 14: I think for portability we could use `-l:libpng.a`, right?

xavery: That would also require adding `/usr/local/lib` to the library path via `-L`, as this path isn't searched by default. Seeing how the value of this variable is used as input to `target_link_libraries`, I don't think there's an easy way of doing that. Doing `-L/usr/local/lib` doesn't seem to work, and adding these flags separately (i.e. `-L;/usr/local/lib/`) leads cmake to think that the path is actually a path to a library which doesn't exist and is removed from the command line, ending up with `-L -l:libpng.a` in the linker invocation, which is just incorrect.

Besides, adding `-L` might lead to more trouble than necessary, as according to `ld`'s manpage "Directories specified on the command line are searched before the default directories. All `-L` options apply to all `-l` options, regardless of the order in which the options appear." which just sounds like trouble.

I'd rather stick to this, as I think just specifying the exact path to a specific library is a cleaner approach here.

Issue #490, opened 2021-11-10
1 ms of difference


vandenk:
Hello,

it's weird I tested with the 2021-11-07-02-04-56 and the 2021-10-13.
When I shift (in -) an audio track (-834 ms) when I re-open there is always 1 ms of difference
![TS1](https://user-images.githubusercontent.com/11425010/141188084-9c0c6a13-8108-4d0c-9df7-90d2ad504a4a.jpg)
Results : 
![TS2](https://user-images.githubusercontent.com/11425010/141188102-219510e2-e94e-4197-8599-4c61844c8fe5.jpg)
it does it with all files (in -)

vandenk:
is this 1ms difference normal ?

PR #508, opened 2021-12-19
[Bug] Fix delay rounding

jcdr428:
When delay is negative, the integer truncating needs to be done after adding -0.5, not +0.5.


jcdr428:
Fixes issue #490 .

Issue #567, opened 2022-03-30
Add more languages


CAPRICORNX10X:
for example I would like the Hebrew language

justdan96:
If you would like to add Hebrew please have a look at https://github.com/justdan96/tsMuxer/tree/master/tsMuxerGUI/translations to see how others have added a new language. We rely on volunteers to provide translations.

CAPRICORNX10X:
I just want to change the caption of the subtitles.
Instead of English - the word in the Hebrew language
Instead of Hebrew - the word in the Hebrew language
![‏‏צילום מסך777 (15)](https://user-images.githubusercontent.com/102700898/162479508-6f5f5d97-c24c-45fd-9e4f-556da21393cc.png)

justdan96:
We have a full Hebrew translation as #581 - but more translations are always welcome!

PR #568, opened 2022-03-30
add language

CAPRICORNX10X:
, for example I would like the Hebrew language


justdan96:
Closing as this should be an issue, and it is in #567

PR #569, opened 2022-03-31
Fix Nightly Release Uploads

justdan96:
A long overdue change this one. GitHub in more recent times has seen more frequent ECONNRESET errors when uploading release artefacts. GitHub has not resolved these issues (that may be ratelimiting for all we know) so we need to replace deprecated actions/create-release and actions/upload-release-asset with ncipollo/release-action. This new version can handle retrying, which seems to be the only way to work around this issue.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,7 +10,7 @@ jobs:
     
     outputs:
       status: ${{ steps.check.outputs.status }}
-      upload_url: ${{ steps.create.outputs.upload_url }}
+      version: ${{ steps.check.outputs.version }}
     
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


justdan96:
I was hoping to do some more checking offline but there doesn't appear to be a way to validate GitHub Actions. I'll merge this one and then check for issues in the morning.

xavery:
Unfortunately it looks like this still occurs even with `ncipollo/release-action` :facepalm: 

Tonight's nightly failed to upload `lnx.zip` - https://github.com/justdan96/tsMuxer/actions/runs/2080693953 . To make matters worse, the build is reported as successful so you don't get a notification and don't know that the release is missing an artifact unless you actually have a look at the release page.

Apparently it's been an issue for some time globally , for example https://github.com/actions/upload-artifact/issues/116 was created in August 2020 , while https://github.com/github/feedback/discussions/9121 has been unanswered since December 2021.

People have apparently had some success with just running the GitHub Client (`gh`) to upload releases, but I can't now find the ticket where this was described. Perhaps doing that plus implementing some kind of back-off in case of failure finally ends this.

justdan96:
I might just fork ncipollo/release-action to add more retries and more time delays in between. I am wondering if consolidating all the files to be uploaded from the Docker into one step might improve things but I'm not sure.

justdan96:
Artefacts were all uploaded successfully last night, let's monitor this for a week to see if we get any more failures and retries.

PR #571, opened 2022-04-02
Move all build/rebuild scripts to "scripts" subdirectory

xavery:
I never really liked all the build scripts being right there in the root directory. This commit moves them all to `scripts` and updates the references.

The scripts themselves don't need changing since all usage information still instructs to launch them from the repository root : if there are any relative paths used inside the scripts, they stay the same.


xavery:
@justdan96 Can you have a look in case I missed a reference somewhere?

justdan96:
Change makes total sense to me, I'll review it in detail later today

justdan96 (approved):
think we've caught all the references

PR #585, opened 2022-04-26
Check for std::filesystem and Correct Linker Flags

justdan96:
This is a first step to supporting std::filesystem, std::experimental::filesystem, ghc::filesystem and boost::filesystem. For now we just check that we can correctly use std::filesystem and we try to figure out the correct linker flags.

### DIFF
--- a/tsMuxer/CMakeLists.txt
+++ b/tsMuxer/CMakeLists.txt
@@ -141,6 +141,11 @@ string(CONFIGURE [[
   }
 ]] GHC_FS_CODE @ONLY)
 
+# required for MacOS, osxcross and LLVM Clang
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(CMAKE_CXX_FLAGS "-std=c++17")
+endif()
+
 # check that we have the standard filesystem installed
 check_cxx_source_compiles("${STD_FS_CODE}" STD_FS_NO_LIB_NEEDED)
 set(CMAKE_REQUIRED_LIBRARIES "stdc++fs")
--- a/tsMuxer/CMakeLists.txt
+++ b/tsMuxer/CMakeLists.txt
@@ -116,6 +116,33 @@ if(TSMUXER_STATIC_BUILD)
   endif()
 endif()
 
+# we can try compiling this small C++ program to test std::filesystem is working
+string(CONFIGURE [[
+  #include <filesystem>
+  int main( int argc, char ** argv ) {
+    std::filesystem::path p( argv[ 0 ] );
+    return p.string().length();
+  }
+]] STD_FS_CODE @ONLY)
+
+# check that we have the standard filesystem installed
+check_cxx_source_compiles("${STD_FS_CODE}" STD_FS_NO_LIB_NEEDED)
+set(CMAKE_REQUIRED_LIBRARIES "stdc++fs")
+check_cxx_source_compiles("${STD_FS_CODE}" STD_FS_NEEDS_STDCXXFS)
+set(CMAKE_REQUIRED_LIBRARIES "c++fs")
+check_cxx_source_compiles("${STD_FS_CODE}"  STD_FS_NEEDS_CXXFS)
+unset(CMAKE_REQUIRED_LIBRARIES)
+
+if( ${STD_FS_NO_LIB_NEEDED} )
+  set( FS_LIBRARIES "" )
+elseif( ${STD_FS_NEEDS_STDCXXFS} )
+  set( FS_LIBRARIES stdc++fs )
+elseif( ${STD_FS_NEEDS_CXXFS} )
+  set( FS_LIBRARIES c++fs )
+else()
+  message( FATAL_ERROR "Unknown compiler - C++17 filesystem library missing" )
+endif()
+
 if (WIN32)
   target_sources(tsmuxer PRIVATE osdep/textSubtitlesRenderWin32.cpp)
   target_link_libraries(tsmuxer gdiplus)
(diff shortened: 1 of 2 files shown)


justdan96:
Clearly something else is going wrong if the Mac native build fails at the CMake stage - I'll mark this as WIP.

justdan96:
This one has grown a bit over what I thought it would take to solve these issues. It does look this is necessary though, as I plan to update the Docker container to only use Ubuntu 18.04 and that would require these changes. @xavery could you please cast an eye over these and suggest if they look okay?

tsMuxer/CMakeLists.txt
```diff
@@ -116,6 +116,76 @@ if(TSMUXER_STATIC_BUILD)
...
+  }
+]] EXP_FS_CODE @ONLY)
+
+string(CONFIGURE [[
+  #include <ghc/filesystem.hpp>
+  int main( int argc, char ** argv ) {
+    ghc::filesystem::path p( argv[ 0 ] );
+    return p.string().length();
+  }
+]] GHC_FS_CODE @ONLY)
```
xavery, line 142: The `CONFIGURE` step isn't needed since there are no `@`-tokens in these strings. You can just use ordinary `set` to initialise these variables : 
```
set(STD_FS_CODE [[
  #include <filesystem>
  int main( int argc, char ** argv ) {
    std::filesystem::path p( argv[ 0 ] );
    return p.string().length();
  }
]])
```

Maybe it's not a big difference, but I find it significantly less confusing.

Using `string(CONFIGURE)` here would be possible if there was one "template" variable which was then "configured" for each variant of `fs` ... something like

```
set(TEMPLATE_FS_CODE [[
  #include <@HEADERFILE>
  using fs = @NAMESPACE;
  int main( int argc, char ** argv ) {
    fs::path p( argv[ 0 ] );
    return p.string().length();
  }
]])
```

Then you could call `string(CONFIGURE)` for each of the combinations to get the final source code to compile, setting `HEADERFILE` and `NAMESPACE` before each call. I like the explicit version much more, this feels like overcomplicating things for no good reason.

tsMuxer/CMakeLists.txt
```diff
@@ -116,6 +116,76 @@ if(TSMUXER_STATIC_BUILD)
...
+
+# required for MacOS, osxcross and LLVM Clang
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(CMAKE_CXX_FLAGS "-std=c++17")
```
xavery, line 146: This doesn't feel right. The main CMake file already has
```
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
set(CMAKE_CXX_EXTENSIONS FALSE)
```
Which should result in `-std=c++17` being there already. I'll investigate and report back - I suspect `std=c++17` isn't the cause of trouble here.

tsMuxer/osdep/textSubtitlesRenderFT.cpp
```diff
@@ -27,7 +27,20 @@ const static char FONT_ROOT[] = "/System/Library/Fonts/";
...
+
+#ifdef FS_EXP
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#endif
+
+#ifdef FS_GHC
+#include <ghc/filesystem.hpp>
+namespace fs = ghc::filesystem;
+#endif
```
xavery, line 43: Can be replaced with a little more (IMHO) straightforward if/else tree : 
```suggestion
#if defined(FS_EXP)
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
#elif defined(FS_GHC)
#include <ghc/filesystem.hpp>
namespace fs = ghc::filesystem;
#else
#include <filesystem>
namespace fs = std::filesystem;
#endif
```

This also won't break horribly if both `FS_EXP` and `FS_GHC` were defined, for whatever reason.

xavery:
What a mess ... wasn't `std::filesystem` supposed to make life easier? :rofl: 

The only reason why `std::filesystem` usage was introduced in this file was to handle symlinks to font files, in 1d50191c0a3c40475cd1684654be4bb148eedaa5 . I don't think anybody expected it to be such a headache, requiring extra 100-some lines in cmake just to build, back then.

The FreeType rendered is used only on Linux and Mac anyway, and I don't think anybody ever attempted using it on Windows. Taking this into account ... wouldn't it be possible to just use `realpath` here and avoid `std::filesystem` problems altogether? Have to admit I haven't tried using it under Mac yet.

tsMuxer/CMakeLists.txt
xavery: Okay. For some reason `-std=c++17` doesn't reach the complier command line when compiling on Catalina. According to [the documentation](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html), the variable `CMAKE_REQUIRED_FLAGS` can be set to force including a flag, which feels less messy than modifying `CMAKE_CXX_FLAGS` directly. Worked for me.

xavery:
The changes replacing `std::filesystem` with `realpath` are in #588 . It looks like they build okay on both Linux and Mac. I don't think it makes a lot of sense to stick to `std::filesystem` seeing how we only use it in this one place, but I understand if you disagree and choose to merge this instead, seeing how you've done the hard part already.

justdan96:
It did feel like creating a Rube Goldberg to crack an egg, thanks for looking over it!

justdan96:
#588 obsoletes these changes

PR #588, opened 2022-04-27
replace std::filesystem usage with realpath()

xavery:


justdan96:
I like this solution a lot better - I'll check everything works in the Docker container and then we should be okay to merge!

justdan96:
All the builds are looking good - I'll close the other PR and we can merge this one.

PR #594, opened 2022-05-03
Fix broken link to compiling documentation

quietvoid:
Title


justdan96:
Quite right - thanks for the contribution!

PR #597, opened 2022-05-08
Tab "About" - Small correction Translation Language German

Ser-vie:
I have corrected a translation error in the "about_de" file.
I had not translated the first sentence, at the tab "About".
Here is the new file
[about_de.zip](https://github.com/justdan96/tsMuxer/files/8647877/about_de.zip)
.


justdan96:
I think some commits need to be squashed, will have a look this week

xavery:
Committed in 24ae93589c64fe9a7b11559801a2fa4a3677c928 .

PR #664, opened 2023-01-24
Added Spanish translation!

Julvenzor:
Hi!

I am a Spanish writer and linguist. I like your program and I have decided to translate it into Spanish. I have created the two corresponding files for the UI and the embedded HTML.

Thank you very much!

### DIFF
--- a/tsMuxerGUI/translations/about_es.html
+++ b/tsMuxerGUI/translations/about_es.html
@@ -1,0 +1,64 @@
+<!-- translators : please *don't* change the formatting of this file, as Qt is very sensitive about it
+  and the translated file will most definitely be rendered differently if you choose to reindent it.
+  also, remember to save the file as UTF-8. -->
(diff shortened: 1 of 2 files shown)


jcdr428:
Hola @Julvenzor, thank you very much for the work! 
Can you please change in tsMuxergui_es.ts all `<translation type="unfinished">` to `<translation>` where the related translation is complete. The QT principle is that the GUI still shows the english text until the translation is marked as finished.

Julvenzor:
Done! When reviewing it, I have corrected a small error: in Spanish, a comma is used, instead of a period, for decimals. Feel free to let me know in the future if there is a mistake or more strings to translate.

By the way, I use the Linux version (appimage). I used the Windows version years ago and, then, I had to keep one partition to be able to rip movies. Thank you very much for making it free (libre) software. Better late than never! (like my translation...)

Regards.

PR #667, opened 2023-01-27
Added Spanish language to language selector

Julvenzor:
Hi again,

Although the translation is done, it is not yet visible. I think it will not appear in the "language select combo box" if it is not indicated in this file. I'm not a programmer yet (I'm learning), but my logic tells me that this file will look for the compiled version of the translation. Such compiled version should be created automatically during build, right?

I don't know if the program code has any way to detect the default language of the user's system.

Regards.


Julvenzor:
Oops, it's wrong. I know where it fails, I think it can be fixed by also modifying the "tsmuxerwindow.cpp" file.

jcdr428:
@Julvenzor you can try your own changes by creating your fork, and modifying the related script for the creation of both tsMuxer and tsMuxerGUI exe.

PR #686, opened 2023-03-08
Fix PTS error

### LINKED ISSUE #677: Incorrect PTS value
The video ES to mux has frame rate of 30000/1001 fps. 
The PTS difference between two consecutive frames (in presentation order) should be 3003 tick in 90k Hz. 
However, we found that one 3002 tick PTS difference in the muxed TS. 

Please find the file to reproduce this issue:
https://drive.google.com/drive/folders/11TfUSrX-TOs-dLJP0duDKAJD5azy2Enb?usp=sharing

To generate the muxed TS, run:
`tsMuxeR 1.meta 1.ts`

In 1.ts, strange PTS diff with 3002 is found at frame 8334 (0 base):
```
FrammeIdx: 8333	byte_pos: 424428 PTS: 25149899
FrammeIdx: 8334	byte_pos: 424721 PTS: 25152901
PTS_diff: 3002
```

moveman:
This is to fix #677 

1. Change to use 196 x 27000000 Hz clock frequency as the unit of PTS instead of using nanosecond. With this unit, frame duration of video frame of common frame rate can be expressed in integer and calculation will be precise. Sample duration of common audio codec of common audio sample rate can also be expressed in integer. 
2. Change the correctFps() function to support more frame rate. This function is needed to ensure the m_pcrIncPerFrame is integer and has no precious loss for common frame rate.

Sample output files before the change and after the change are put here, along with the PTS analysis excel:
https://drive.google.com/drive/folders/1fCWmNde66EUb5VE8GEyh_ZneqBTStXGa?usp=sharing

There are many code used the term Nano which does not match the change. We need to come up with a new term.


jcdr428:
@moveman 
1. Please fix clang format check errors https://github.com/justdan96/tsMuxer/actions/runs/4370087634/jobs/7648701565
2. Please see whether the 0.5 roundings to the nearest are still needed with unusual fps (e.g. animes 8 to 15 fps).

jcdr428 (changes requested):
@moveman
1. Please fix clang format check errors https://github.com/justdan96/tsMuxer/actions/runs/4370087634/jobs/7648701565
2. Please see whether the 0.5 roundings to the nearest are still needed with unusual fps (e.g. animes 8 to 15 fps).

jcdr428:
@moveman I've kept the rounding to nearest as this should not affect your purpose. Please confirm whether this solves the issue.

ValZapod:
>Please confirm whether this solves the issue.

It solves the issue here:

[Parsed_vfrdet_0 @ 000002061ee98b80] VFR:0.000000 (0/18419)

moveman:
Hi @jcdr428, the rounding has to be removed. Consider the original code:
`
        newPCR = (uint64_t)((m_endStreamDTS - m_minDts) / INT_FREQ_TO_TS_FREQ + 0.5 + m_fixed_pcr_offset);
`
The type of INT_FREQ_TO_TS_FREQ is changed from double to int64_t, thus, `(m_endStreamDTS - m_minDts) / INT_FREQ_TO_TS_FREQ` will become an integer. The `+ 0.5` will become meaningless. 

if `newPCR = (uint64_t)((m_endStreamDTS - m_minDts) / INT_FREQ_TO_TS_FREQ + 0.5 + m_fixed_pcr_offset);` can work well with the PTS error fix, I think changing it to `newPCR = (m_endStreamDTS - m_minDts) / INT_FREQ_TO_TS_FREQ + m_fixed_pcr_offset;` should not cause any problem.

moveman:
Many thanks @jcdr428.

PR #688, opened 2023-03-08
PAT PMT should better be placed in the beginning of the TS

moveman:
Player behaves better if PAT PMT are placed in the beginning of the TS. You can try playing the video here using VLC:
https://drive.google.com/drive/folders/1KWFEYB2pt7ON9xRVikciioYirlQldnqI?usp=sharing

before.ts: mux-ed output before this change
after.ts: mux-ed output after this change

This pull request included some changes from the another pull request... so let's consider this pull request later.

### DIFF
--- a/tsMuxer/tsMuxer.cpp
+++ b/tsMuxer/tsMuxer.cpp
@@ -981,6 +981,13 @@ bool TSMuxer::muxPacket(AVPacket& avPacket)
     if (tsIndex == 0)
         THROW(ERR_TS_COMMON, "Unknown track number " << avPacket.stream_index);
 
+    auto newPCR = (int64_t)((avPacket.dts - m_minDts) / INT_FREQ_TO_TS_FREQ + 0.5 + m_fixed_pcr_offset);
+    if (m_lastPCR == -1)
+    {
+        writePATPMT(newPCR, true);
+        writePCR(newPCR);
+    }
+
     bool newPES = false;
     if (avPacket.dts != m_streamInfo[tsIndex].m_dts || avPacket.pts != m_streamInfo[tsIndex].m_pts ||
         tsIndex != m_lastTSIndex || avPacket.flags & AVPacket::FORCE_NEW_FRAME)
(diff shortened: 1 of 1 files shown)


jcdr428:
@moveman generally seems good to me -we'll have to find another naming for the nano things.
Only the comment on the 0.5 roundings: IMO these ones are still needed when using unusual fps e.g. some animes use 8 to 15 fps.

moveman:
Hi @jcdr428, we don't have good solution for unusual fps. But given that 1 PTS error is an existing problem, maybe it should be okay we don't fix it for 8 to 15fps? What problem we will encounter if we don't have 0.5 roundings?

jcdr428:
@moveman is it a problem for you to leave the rounding to nearest rather than rounddown ?

ValZapod:
>You can try playing the video here using VLC:

You should not. VLC is very buggy. Use mpv and ffplay instead to test such things.

Besides what is a real problem is all of those 

[ffmpeg/video] h264: co located POCs unavailable

But that cannot be fixed (ffmpeg's ts has it too) without ffmpeg.exe before.ts -c copy fnafcva.mp4

jcdr428:
@moveman can you please propose a new commit taking into account that your two previous  commits were merged.

ValZapod:
>Besides what is a real problem is all of those
>[ffmpeg/video] h264: co located POCs unavailable
>every other time you seek.

Should have provided some references: https://trac.ffmpeg.org/ticket/9412, in particular it appears that [this patch](https://patchwork.ffmpeg.org/project/ffmpeg/patch/20190831101039.2179-1-fumoboy007@me.com/) series may help with that seeking which is more complex in naive MPEG-TS (and Microsoft did manage to write correct code for seek in TS) compared to much more nice and reference containers like mp4 or even (not really, actually) mkv.

moveman:
hi @jcdr428 @ValZapod, I have cleaned up this commit. Actually we do not rely on VLC. I just want to say that this change could help some players and analyzers. The real problem we suffered is that some functionality of tsduck does not work well if PAT/PMT is not placed in the beginning.

PR #815, opened 2023-12-27
Do not append file

moveman:
In file system like Goofys and MountPoint for S3, we cannot append file. Hence, we cannot use tsMuxeR there. To overcome this, we should not use append.


moveman:
Hi, may I know why we need to close and reopen the output file for writting? Thanks.

jcdr428:
> Hi, may I know why we need to close and reopen the output file for writting? Thanks.

I guess the file is initially in "write" mode only, so needs to be closed and reopened in "write + append" mode.

moveman:
Why not just write all the things in write mode?

在 2024年5月7日週二 01:33，jcdr428 ***@***.***> 寫道：

PR #818, opened 2024-01-08
Bug fixed in iso_writer.cpp

fxg79:
ECMA-167 - 7.2.3 Tag Checksum (RBP 4)
This field shall specify the sum modulo 256 of bytes 0-3 and 5-15 of the tag.


fxg79:
Ok, this is not a bug if we assume that buffer[4] is zero...

PR #833, opened 2024-03-04
PGS demuxing: set DTS equal to PTS for packets with PTS_DTS_flags=0b10

cubicibo:
This is a tiny change in demuxing. I think this is more logical than a sporadically null DTS in the output .SUP file.

- `PTS_DTS_flags=0b10` suggests PTS=DTS, rather than DTS=0.
- To mux back the stream, it is easier to check for PTS=DTS to find back `0b10 == PTS_DTS_flags` candidates rather than look at the segment type and do guess work, as DTS=0 can be a valid value, while PTS and DTS should never be equal in a TS packet.

### DIFF
--- a/tsMuxer/pgsStreamReader.cpp
+++ b/tsMuxer/pgsStreamReader.cpp
@@ -811,10 +811,7 @@ int PGSStreamReader::writeAdditionData(uint8_t* dstBuffer, uint8_t* dstEnd, AVPa
         *dstBuffer++ = 'G';
         auto data = reinterpret_cast<uint32_t*>(dstBuffer);
         *data++ = my_htonl(static_cast<uint32_t>(internalClockToPts(m_lastPTS)));
-        if (m_lastDTS != m_lastPTS)
-            *data = my_htonl(static_cast<uint32_t>(internalClockToPts(m_lastDTS)));
-        else
-            *data = 0;
+        *data = my_htonl(static_cast<uint32_t>(internalClockToPts(m_lastDTS)));
         return 10;
     }
     return 0;


jcdr428:
@cubicibo not sure about this: in the below links, dts should be zero. If you can find a reference to a more official source...

https://blog.thescorpius.com/index.php/2017/07/15/presentation-graphic-stream-sup-files-bluray-subtitle-format/

https://video.stackexchange.com/questions/16261/split-pgs-subtitles-sup-file

jcdr428:
@cubicibo not sure about this: in the below links, dts should be zero. If you can find a reference to a more official source...

https://blog.thescorpius.com/index.php/2017/07/15/presentation-graphic-stream-sup-files-bluray-subtitle-format/

https://video.stackexchange.com/questions/16261/split-pgs-subtitles-sup-file

cubicibo:
Hi @jcdr428, unfortunately this is a widespread misconception originating from DGDemux/eac3to suites. Many files were produced with these programs that always set DTS to zero, and people got the impression the field is never used.

However, when drmpeg designed the 'PG' SUP file format, he made sure to convey both PTS and DTS (truncated from 33 to 32 bits). Since the DTS is irrelevant for software players, it has been discarded by some tools like DGDemux.
If the SUP file is muxed back to a transport stream, the incorrect DTS value will cause issue to some decoders. It can range from PTS-DTS difference being larger than one second, memory overruns in the PG decoder (too many PCS buffered at once, overwriting ODS when it should be read-only). Or not enough decoding for some segments time if tsMuxer uses PTS for scheduling when DTS is zero.

When muxing, tsMuxer and other professional muxers uses the "not permitted" case of PTS=DTS to know when a PES packet should have PTS_DTS_Flags equal to 0b10 or 0b11: https://github.com/justdan96/tsMuxer/blob/94cafe7244213870aaab37035c827ef839a15929/tsMuxer/tsMuxer.cpp#L609C1-L612C51

This proposed code change makes tsMuxer muxing and demuxing operations bijective for PG streams.

jcdr428:
Ok, I've pushed the commit.

PR #875, opened 2024-05-09
Fix race condtion that cause corrupted tail of output TS

moveman:
We found that occasioinally the tail of the TS output is corrupted. After some debugging and guessing, the problem should be caused by this code
```
m_nothingToExecute = m_writeQueue.empty();
```
This code is not atomic. It is equivalent to this:
```
bool flag = m_writeQueue.empty();
m_nothingToExecute = flag;
```
Other threads could jump in after the first line and before the second line. 
This includes BufferedFileWriter::addWriterData(). When addWriteData() is indeed be called, 
the situation will be like this:
```
bool flag = m_writeQueue.empty(); // (true)

  m_nothingToExecute = false;
  m_writeQueue.push(data);

m_nothingToExecute = flag; // (true)
```
writeQueue is inserted a data and is NOT empty; however, m_nothingToExecute is set to true by the final line. 
That will cause waitForWriting() to return and then close() is called before the remaining byte is flushed to the output.

### DIFF
--- a/libmediation/containers/safequeue.h
+++ b/libmediation/containers/safequeue.h
@@ -50,6 +50,15 @@ class SafeQueue
         return val;
     }
 
+    T peek()
+    {
+        std::lock_guard<std::mutex> lk(m_mtx);
+
+        T val = m_queue.front();
+
+        return val;
+    }
+
    private:
     mutable std::mutex m_mtx;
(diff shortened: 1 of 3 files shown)


ValeZAA:
Can you provide a sample source file and output: before and after?

